PR #1, opened 2017-12-07
Setup CircleCI, Docker, Makefile, Pylint, and Accounting app.

UmanShahzad:
This just adds a skeletal version of the app with some base frameworks (mostly) setup.

**Testing Instructions**:

1. Follow the README.md for developmental setup and be sure that it 'just works'.


UmanShahzad:
@itsjeyd Thanks!

requirements/base.txt
```diff
@@ -0,0 +1,19 @@
+astroid==1.5.3
+django==1.11.8
```
antoviaque: @UmanShahzad Django 2 has been released: https://docs.djangoproject.com/en/2.0/releases/2.0/

Can you use the latest version of all the dependencies, to not write against outdated versions?

UmanShahzad: @antoviaque All dependencies are the latest, yeah. I didn't want to use Django 2.0 just yet though because I worry if some of the Django apps I'm using won't be Django 2.0 compatible. Since we're already using Python 3.6 for this, upgrading to Django 2.0 would be trivial if we wanted to do it since that's the main change -- deprecating Python 2 support.

PR #5, opened 2017-12-11
Implement authentication and registration back-end

UmanShahzad:
## Registration

- Get into the proper virtual environment.
- Do `pip install -r requirements.txt`.
- Do `cp .environ/.env.dev .env`.
- Do `docker-compose up -d`.
- When the docker containers all start up, do `make dev.web.logs`.
- When the containers are running, do `make migrate`.
- Register a new user through the API. For example:

```bash
curl -X POST -H 'content-type: application/json' -d '{"username": "mslm", "password": "password123", "last_name": "Shahzad", "first_name": "Uman", "email": "uman@opencraft.com"}' http://localhost:1786/registration/register/
```

- You will see a verification link in the logs. Follow it and ensure the payload you receive makes sense and that you are thus verified.

## Authentication

In order to see your user details (i.e. 'log in' from a client perspective), you need to authenticate using tokens.

- Obtain a token from http://localhost:1786/auth/token/obtain/. For example:

```bash
curl -X POST -d 'username={username}&password={password}' http://localhost:1786/auth/token/obtain/
```

- Copy the token you get back.
- See the details using the token. For example:

```bash
curl -X GET -H 'Authorization: Token {token}' http://localhost:1786/auth/user/{username}/
```

- See that you get back a bunch of user details (account will be `null` for now).
- Change the token in the authorization header slightly, and see that you are denied access.

(Optional)

- Play around with revoking/verifying tokens using the http://localhost:1786/auth/token/revoke/ and http://localhost:1786/auth/token/verify/ endpoints, respectively. Both require a `POST` request with a `token` and a `username`. For example:

```bash
curl -X POST -d 'token={token}&username={username}' http://localhost:1786/auth/token/verify/
```

## Account Details

Although it's trivial, I'm low on time so I have not written an endpoint for creating new Account models through the REST framework. So for now, create them in the admin using arbitrary details. Give as much (fake) information as possible. (I will add the endpoints in a later PR).

- You can't make users administrators through the API, so do

```bash
$ make shell
>>> user = User.objects.get()
>>> user.is_staff = True
>>> user.save()
>>> CTRL+D
```

- Login to http://localhost:1786/admin and make the `Account` model for your user. Also make a `BankAccount` for the resulting `Account` as a separate but extra piece.

We're staff now so tokens are irrelevant. Verify endpoints directly in the browser view now.

- Go to http://localhost:1786/account/{username}/ and see that the info matches what you'd expect for account details.
- Go to http://localhost:1786/account/address/{username}/ and see that the info matches what you'd expect for address details.

## Bank Details

Bank details require UUIDs to retrieve. Get them from the command line as I believe the admin doesn't have them showing.

- Go to http://localhost:1786/bank/{uuid}/ and see that the info matches what you'd expect for the bank details (not the bank *account*).
- Go to http://localhost:1786/bank/account/{uuid}/ and see that the info matches what you'd expect for the bank *account* details.

## Hourly Rates

Hourly rates are a more complex feature. Any user account can be a provider or a client related to another user account which is the opposite (i.e. they're the client if you are the provider) and the relation between you would be the hourly rate in an arbitrary currency. You can even bill yourself.

For now, we must create these through the Django admin.

- Go to http://localhost:1786/admin/account/hourlyrate/ and create a new hourly rate between 2 accounts -- this may require you to make a new one. For example, make the *real you* the provider, and OpenCraft the client, so that you charge OpenCraft some EUR.
- Go to http://localhost:1786/account/rate/{provider_username}/{client_username}/ and ensure that you get what you expect for the hourly rate between the provider and client. Example link:

http://localhost:1786/account/rate/mslm/OpenCraft

- Create *another* hourly rate between the same provider and client (obviously not a very real scenario, but still).
- Go to the same link and see that you now can see a list of hourly rates between the two.


itsjeyd:
@UmanShahzad Just to make sure, you don't need me to make a pass over this just yet, right?

UmanShahzad:
@itsjeyd Yep, still working on it atm.

UmanShahzad:
@itsjeyd Ready for review!

accounting/bank/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+                ('iban', localflavor.generic.models.IBANField(blank=True, help_text='The unique International Bank Account Number of the provider.', include_countries=None, max_length=34, null=True, use_nordea_extensions=False)),
+                ('bic', localflavor.generic.models.BICField(blank=True, help_text='The 11-character SWIFT / BIC (Business Identifier Code) code used to identify a bank or financial institution globally.', max_length=11, null=True)),
+                ('abn', localflavor.au.models.AUBusinessNumberField(blank=True, help_text='The 11-character Australian Business Number used to identify business entities in Australia.', max_length=11, null=True)),
+                ('bsb', accounting.bank.fields.AUBankStateBranchField(blank=True, help_text='The 6-character Bank State Branch code used as a bank identifier in Australia.', max_length=6, null=True)),
+                ('vat', vies.models.VATINField(blank=True, help_text='The value added tax identification number used in and required by some countries. This can also be used to store a general sales tax (GST) identification number.', max_length=14, null=True)),
```
itsjeyd: @UmanShahzad I like how you capitalized words making up the IBAN and BIC acronyms above. It would be nice to do that for VAT as well ("The Value Added Tax identification number...").

accounting/bank/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+                ('account_number', models.CharField(blank=True, help_text='The bank account number used to help identify the account. Required for only some countries.', max_length=30, null=True)),
+                ('routing_number', models.CharField(blank=True, help_text='The bank account routing number used to help identify the account. Required for only some countries.', max_length=30, null=True)),
+                ('bank', models.ForeignKey(help_text='The bank in which this bank account is opened.', on_delete=django.db.models.deletion.CASCADE, related_name='bank_accounts', to='bank.Bank')),
```
itsjeyd: @UmanShahzad This sounds like it's the accounting app's business to open bank accounts :) I'd probably change this to "The bank to which this bank account belongs."

accounting/registration/views.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class RegistrationView(CreateAPIView):
+    """
+    View to register a user and send a verification email.
+    """
+
+    serializer_class = CreateUserSerializer
+
+    def perform_create(self, serializer):
```
itsjeyd: @UmanShahzad This is missing a docstring.

accounting/registration/views.py
```diff
@@ -0,0 +1,86 @@
...
+        email = None
+        verified = expired = False
+        response_status = status.HTTP_200_OK
+        try:
+            email = EmailAddress.objects.confirm(code).email
+        except EmailAddress.DoesNotExist:
+            response_status = status.HTTP_404_NOT_FOUND
+        except EmailConfirmationExpired:
+            expired = True
+        else:
```
itsjeyd: @UmanShahzad YAY for `try-except-else` 😉

itsjeyd:
@UmanShahzad Ah, one more thing: The README still says to do

```
(accounting) $ pip install -r requirements/dev.txt
```

as part of the setup process.

That should probably be updated to match what you put in the PR description.

.environ/.env.dev
```diff
@@ -1,6 +1,12 @@
-# Copy this to `../.env` (to be at the root) so it's picked up.
+# Copy this to `../.env` (to be at the root) so it's picked up by Django settings.
 # Make sure you have the docker containers running while developing.
 # See the README for understanding how to get your developmental environment setup.
+#
+# Note that although Docker also looks at `.env` files for picking up environmental
+# settings, values present in the docker environment at runtime, e.g. through `docker-compose.yml`,
+# take precedence over these. `docker-compose.yml` has sane defaults. You still need these for the
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

README.md
```diff
@@ -18,7 +18,7 @@ In fact, to run tests or basically anything with Django management commands, you
 
 ### Environmental Settings
 
-A sample development environment settings file is in `.environ/.env.dev`. You can copy this over to your root as `.env` to use if you'd like to run `make tests` outside of a Docker container, for example.
+A sample development environment settings file is in `.environ/.env.dev`. You can copy this over to your root as `.env` to use if you'd like to run `make tests` outside of a Docker container, for example. It can also be used to override the default environmental variables passed into docker in `docker-compose.yml`.
```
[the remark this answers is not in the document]
UmanShahzad: Yeah this is confusing. Both Docker and Django give `.env` priority, so that's what I should have said -- `.env` has priority over `docker-compose.yml`. I will change the statement in the earlier comment.

accounting/account/admin.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
+# Copyright (C) 2015-2017 OpenCraft <contact@opencraft.com>
```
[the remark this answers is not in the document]
UmanShahzad: I don't really know, tbh. @antoviaque ?

accounting/account/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+            },
+        ),
+        migrations.CreateModel(
+            name='Address',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('country', django_countries.fields.CountryField(help_text='The country associated with this user account.', max_length=2)),
+                ('address_line1', models.CharField(help_text='The first address line used to appear on accounting documents, i.e. invoices.', max_length=128)),
+                ('address_line2', models.CharField(blank=True, help_text='Additional line for extending an address.', max_length=128, null=True)),
+                ('zipcode', models.CharField(help_text='A 5-digit or ZIP+4 zipcode. Example: 12345, 12345-6789.', max_length=10)),
```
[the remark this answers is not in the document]
UmanShahzad: Oh my lack of standards. An engineer's nightmare.

accounting/account/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
...SP', 'South Sudanese Pound'), ('LKR', 'Sri Lanka Rupee'), ('XSU', 'Sucre'), ('SDG', 'Sudanese Pound'), ('SRD', 'Surinam Dollar'), ('SEK', 'Swedish Krona'), ('CHF', 'Swiss Franc'), ('SYP', 'Syrian Pound'), ('BDT', 'Taka'), ('WST', 'Tala'), ('TZS', 'Tanzanian Shilling'), ('KZT', 'Tenge'), ('XXX', 'The codes assigned for transactions where no currency is involved'), ('TTD', 'Trinidad and Tobago Dollar'), ('MNT', 'Tugrik'), ('TND', 'Tunisian Dinar'), ('TRY', 'Turkish Lira'), ('TMT', 'Turkmenistan New Manat'), ('TVD', 'Tuvalu dollar'), ('AED', 'UAE Dirham'), ('XFU', 'UIC-Franc'), ('USD', 'US Dollar'), ('USN', 'US Dollar (Next day)'), ('UGX', 'Uganda Shilling'), ('CLF', 'Unidad de Fomento'), ('COU', 'Unidad de Valor Real'), ('UYI', 'Uruguay Peso en Unidades Indexadas (URUIURUI)'), ('UYU', 'Uruguayan peso'), ('UZS', 'Uzbekistan Sum'), ('VUV', 'Vatu'), ('CHE', 'WIR Euro'), ('CHW', 'WIR Franc'), ('KRW', 'Won'), ('YER', 'Yemeni Rial'), ('JPY', 'Yen'), ('CNY', 'Yuan Renminbi'), ('ZMK', 'Zambian Kwacha'), ('ZMW', 'Zambian Kwacha'), ('ZWD', 'Zimbabwe Dollar A/06'), ('ZWN', 'Zimbabwe dollar A/08'), ('ZWL', 'Zimbabwe dollar A/09'), ('PLN', 'Zloty')], default='EUR', editable=False, max_length=3)),
```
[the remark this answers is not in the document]
UmanShahzad: This is actually from upstream `djmoney`, not us.

accounting/account/models.py
```diff
@@ -0,0 +1,146 @@
...
+class HourlyRate(models.Model):
+    """
+    An hourly rate model that serves as a way to map a provider's hourly rate per client.
+
+    For technical details, see Django documentation, which explains intermediate models like this one:
+    https://docs.djangoproject.com/en/2.0/topics/db/models/#extra-fields-on-many-to-many-relationships
+    """
+
+    # Allows up to 999999.99 of an hourly rate for some currency.
+    # There are some currencies that are quite inflated. Or this provider is just really darn expensive!
```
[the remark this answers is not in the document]
UmanShahzad: :p

accounting/account/models.py
```diff
@@ -0,0 +1,146 @@
...
+    hourly_rate = MoneyField(
+        max_digits=8, decimal_places=2,
+        help_text=_("The hourly rate charged between a provider and client."))
+    provider = models.ForeignKey(
+        Account, models.CASCADE, related_name='provider_hourly_rates',
+        help_text=_("The account linked to the user who gets paid the hourly rate by the client."))
+    client = models.ForeignKey(
+        Account, models.CASCADE, related_name='client_hourly_rates', blank=True, null=True,
+        help_text=_("The account linked to the user who pays the hourly rate to the provider."))
+
```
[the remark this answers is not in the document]
UmanShahzad: Yep, for the reason above; it's from `djmoney`, not us.

accounting/account/models.py
```diff
@@ -0,0 +1,146 @@
...
+        For example:
+            - Expensive Consultant charges Big Enterprise 500 EUR / hour
+            - Lowly Dev charges Unicorn Startup 15 EUR / hour
+        """
+        provider_name = self.provider.business_name or self.provider.user.get_full_name()
+        client_name = self.client.business_name or self.client.user.get_full_name()
+        return '{provider} charges {client} {rate} / hour'.format(
+            provider=provider_name,
+            client=client_name,
+            rate=self.hourly_rate,
```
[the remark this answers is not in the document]
UmanShahzad: Yep it does -- you can check it out in the shell, but I've already seen it and made sure it was this way.

accounting/account/urls.py
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+Account URL Configuration.
+"""
+
+from django.conf.urls import url
+
+from accounting.account import views
+
+USERNAME_REGEX = r'[\w .@_+-]+'
```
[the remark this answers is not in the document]
UmanShahzad: Oh this is copy and paste from I think edx-platform. Will remove it but not entirely sure of the consequences (my brain's a little too fried for regex right now).

accounting/account/views.py
```diff
@@ -0,0 +1,83 @@
...
+        obj = get_object_or_404(queryset, **self.kwargs)
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def retrieve(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset()).filter(**self.kwargs)
+        try:
+            instance = self.get_object()
+        except MultipleObjectsReturned:
+            data = {'hourly_rates': [self.get_serializer(hourly_rate_data).data for hourly_rate_data in queryset]}
```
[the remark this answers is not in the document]
UmanShahzad: You're right, we really shouldn't. I've added `unique_together` for them.

accounting/authentication/urls.py
```diff
@@ -0,0 +1,36 @@
...
+from accounting.authentication.views import ObtainAuthToken, RevokeAuthToken, UserView, VerifyAuthToken
+
+token_urlpatterns = [
+    url(r'^token/obtain/', ObtainAuthToken.as_view()),
+    url(r'^token/revoke/', RevokeAuthToken.as_view()),
+    url(r'^token/verify/', VerifyAuthToken.as_view()),
+]
+
+urlpatterns = [
+    url(r'^user/{username}'.format(username=r'(?P<username>[\w .@_+-]+)'), UserView.as_view()),
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/bank/fields.py
```diff
@@ -0,0 +1,46 @@
...
+Bank fields used by any of the Bank models.
+"""
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class AUBankStateBranchField(models.CharField):
+    """
+    A model field that checks that the value is a valid Australian Bank State Branch (BSB) code.
```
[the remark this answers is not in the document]
UmanShahzad: Sure, I've added some explanation and a link for reference.

accounting/bank/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+            },
+        ),
+        migrations.CreateModel(
+            name='BankAccount',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('uuid', models.UUIDField(default=uuid.uuid4, editable=False, help_text='The universally unique identifier for this bank account.', verbose_name='UUID')),
+                ('currency', djmoney.models.fields.CurrencyField(default='EUR', help_text='The currency expected to be held in this bank account.', max_length=3)),
+                ('type', models.CharField(choices=[('checking', 'Checking Account'), ('savings', 'Savings Account')], help_text='Whether this is a checking or savings account.', max_length=30)),
+                ('iban', localflavor.generic.models.IBANField(blank=True, help_text='The unique International Bank Account Number of the provider.', include_countries=None, max_length=34, null=True, use_nordea_extensions=False)),
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/bank/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Bank',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('uuid', models.UUIDField(default=uuid.uuid4, editable=False, help_text='The universally unique identifier for this bank.', verbose_name='UUID')),
+                ('name', models.CharField(help_text='The official name of the bank.', max_length=80)),
+                ('address', models.ForeignKey(help_text='The address of this user.', on_delete=django.db.models.deletion.CASCADE, related_name='banks', to='account.Address')),
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/bank/migrations/0001_initial.py
UmanShahzad: :+1:

accounting/bank/migrations/0001_initial.py
UmanShahzad: :+1:

accounting/registration/views.py
UmanShahzad: :+1:

accounting/registration/views.py
UmanShahzad: :+1:

UmanShahzad:
@itsjeyd Alrighty, updates are in. I've also gone ahead and added the API endpoints to this PR to allow creating new accounts/addresses/hourly-rates (not banks yet), although you can skip testing them for now (up to you) since we'll likely have to look at them in the next PR(s).

accounting/bank/fields.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad Thanks, looking great!

UmanShahzad:
@itsjeyd Thanks for the review! Good catches and I've made the appropriate fixes. What do you think about the strings in the admin now? Do they read well?

UmanShahzad:
@itsjeyd Awesome, thanks!

Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM python:3.6
-ENV PYTHONUNBUFFERED=1 \
-    POSTGRES_PASSWORD='password'
+
+RUN rm -rf /accounting /venvs
```
antoviaque: We are using docker? That's different from Ocim's use of Vragrant?

accounting/account/admin.py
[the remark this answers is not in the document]
antoviaque: Yup, it needs to be the time the code was written - so unless we reuse code written before, that would be 2017-2018 yes

accounting/account/migrations/0001_initial.py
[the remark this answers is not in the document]
antoviaque: Note that the rate is always in euros - so I assume this is actually a field to express the currency of the recipient account?

accounting/account/models.py
```diff
@@ -0,0 +1,155 @@
...
+    address = models.ForeignKey(
+        Address, models.CASCADE, related_name='accounts',
+        help_text=_("The address of this user."))
+    business_name = models.CharField(
+        max_length=120, blank=True, null=True,
+        help_text=_("The name of the business or company associated with this account."))
+    occupation = models.CharField(
+        max_length=80, blank=True, null=True,
+        help_text=_("The occupation of the user of this account. "
+                    "Required if country is US, CA, or JP."))
```
antoviaque: This comes from Transferwise? It would be good to have a way to not have to update our code every time a requirement from transferwise changes - ie ask for this information dynamically, based on what they send through the API.

accounting/account/models.py
```diff
@@ -0,0 +1,155 @@
...
+
+    For technical details, see Django documentation, which explains intermediate models like this one:
+    https://docs.djangoproject.com/en/2.0/topics/db/models/#extra-fields-on-many-to-many-relationships
+    """
+
+    # Allows up to 999999.99 of an hourly rate for some currency.
+    # There are some currencies that are quite inflated. Or this provider is just really darn expensive!
+    hourly_rate = MoneyField(
+        max_digits=8, decimal_places=2,
+        help_text=_("The hourly rate charged between a provider and client."))
```
antoviaque: Ah, I see, the hourly rates are generic, not just tied to our invoicing within the team, and the account models too. Not sure if we needed that level of abstraction though, since hourly rates only matter for the team itself - for others, we simply have incoming invoices, with a total, we don't care about the rate. It might have generated complexity in the models and the views needed to generate/edit them that was really necessary at this stage imho - a simple numerical field "hourly_rate" on a DeveloperAccount model would have been enough for now?

accounting/bank/fields.py
```diff
@@ -0,0 +1,51 @@
...
+
+    BSBs are 6-digit codes. From https://www.thebsbnumbers.com/:
+        - The First two digits (XX) specify the parent financial institution.
+        - Third digit (Y) specifies the state where the branch is located.
+        - Fourth, fifth and sixth digits (ZZZ) specify the branch location.
+    """
```
antoviaque: As mentioned above and the email thread on all@, I'd like to avoid handling specific use cases for the different countries manually in our code. Can we delegate that part to Transferwise, only collecting the information they ask for through their API, and having one approach that works for all countries? Otherwise we'll need to update the code evey time their is a change in legislation, or whenever someone joins from a country with a set of rules we haven't implemented yet.

accounting/bank/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+class BankAccount(models.Model):
+    """
+    A bank account stores details to help identify a particular user's bank account anywhere in the world.
+
+    Multiple bank accounts can belong to a single user account.
+
+    Because bank account identifiers differ from country to country (although we're moving towards standardizing
+    on IBAN & SWIFT/BIC), we let all bank account identifying details be optional, to let the user pick and choose
+    what makes sense for their case.
```
antoviaque: Again, as mentioned on all@, we need to ask for the fields that Transferwise is going to require, and segregate cases which Transferwise don't handle. Otherwise people aren't going to enter what is needed exactly, and it's going to be a mess every time someone adds or updates their bank info.

Also we need to check that the bank account matches the country of the invoicing entity, so you probably need to create the transferwise recipient at this stage - and store the info there when we can use them, rather than in our DB.

Dockerfile
UmanShahzad: This is just for the development environment, and I don't think that's an issue -- you just have to run `make dev.up` and it's done :) I don't plan to deploy with Docker though.

We can also put a Vagrantfile to work alongside Docker, but I found Docker easier/faster to get started with.

accounting/account/admin.py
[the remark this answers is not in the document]
UmanShahzad: :+1: alrighty, will change it.

accounting/account/models.py
UmanShahzad: I don't understand what you mean here -- the TransferWise API expects this information in a request to, for example, create a new recipient account. I guess you're saying we'd create an account on their side manually? I imagined we'd let that happen through the registration process -- user would enter these details and their recipient account would be made automatically if they are eligible for TransferWise payments.

accounting/account/models.py
UmanShahzad: Yeah this is quite generic, but it didn't complicate the views/serializers that much, actually. If you take a look at them, there's just one place that requires slight customization, otherwise it's the same as the others.

But I think this is a good starting point for handling outgoing invoices with hourly rates. Not entirely sure what I foresee but I don't think this hurts.

accounting/bank/fields.py
UmanShahzad: So there probably isn't going to be a significant enough change in legislation often enough to have that require code changes, and I've designed this so that if we do have to add a special 'code' or something for a new country, it should be trivial to do so (as trivial as just adding the new field to the model; i.e. 20~ min of work). But I don't think even *that* would be required, since I got  these requirements from TransferWise already.

I think the issue we have is I envisioned that recipient accounts would automatically be created through this app's automation, whereas you envisioned that they already would be created and just re-used through the app. This is where I'm confused, and I think our discussions aren't helping my confusion necessarily: where does the automation start and where does it end? What are its boundaries? I feel like we're only figuring this out just now.

accounting/bank/models.py
UmanShahzad: We need this information to generate the invoice with the details required on the invoice. It's not just about the TransferWise flow, but all flows, including invoice generation. We can't generate invoices without having at least some of these fields, which people say they require on their invoices.

accounting/account/admin.py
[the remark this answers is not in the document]
UmanShahzad: I will do this in the next PR I make, btw, for invoices.

accounting/bank/fields.py
UmanShahzad: I'm going to adopt Braden's recommendation and use a JSONField to hold these attributes.

Dockerfile
antoviaque: @UmanShahzad I won't block on this, but we were also targeting having the same environment as for Ocim, so I would have expected this to be discussed at least. Are there other differences between the two, aside from the dependencies versions which would be more recent here?

accounting/account/models.py
antoviaque: As mentioned elsewhere, for things that Transferwise require, this would be dynamic fields and likely only needs to be stored in Transferwise. If we were to store it on our side eventually, that would go to the JSON field.

accounting/account/models.py
antoviaque: OK, sounds good then, that seem reasonable.

accounting/bank/fields.py
antoviaque: 👍

accounting/bank/models.py
antoviaque: Why does that allows to more easily generate the invoices? They don't need the bank account details, this is only for transfers. Before duplicating info between two places (Transferwise and our DB), and thus having to handle a layer of cache, I would like to understand why this would be helpful. It's easier to just have one reference, Transferwise's API?

Dockerfile
UmanShahzad: I don't think there are other differences, no. But it's possible I'm forgetting something.

But this is a development environment thing, and I would say it should be up to the developer's discretion to choose one? I don't know how `vagrant up` vs. `make dev.up` affects the actual product -- except that the latter allows you to work faster on it.

accounting/account/models.py
UmanShahzad: Just to be sure -- are you saying that the business name / occupation should be in a JSONField?

accounting/bank/models.py
UmanShahzad: So I don't have a particularly strong argument, just a bunch of little points:

1. Not getting completely tied to a 3rd party; this has a bunch of potential consequences of its own
1. Rate limits since we'd require burst calls; this led to issues even for Ocim with the GitHub API
1. Easier to get started for this first iteration; since I won't have time to make a TransferWise API client and work with it in the first iteration, this info will be manually entered, so it'll go in the DB

UmanShahzad: > If we ever want to change provider we could retrieve the data at that point, through the same API. We would likely need to edit it anyway, since another provider would likely require slightly different information (international vs local accounts, different justification, etc.). I'm not too worried about this at this stage.

:+1: 

> When do we need to do burst calls? Can't the recipient id be used in place of the banking information in API calls to Transferwise?

I imagined we'd do burst calls to the TransferWise API for recipient details using the recipient ID every month. We need to do it, right, if we're not storing that info? And even if we store it, if we need to update it we'd do that on TransferWise, still requiring API calls. I have not tested their API yet, but if we have a big enough team, I wonder if they have burst limits that'll cause trouble. Maybe I'm misunderstanding something in your vision here.

> I don't think you need to implement the Transferwise API at this stage - a field could just store the transferwise recipient id, entered manually.

But then where would we get bank details from, if we don't implement the API client? I see now that you said we might not need bank details on the invoice, but I wonder if some people require this for their own invoices.

UmanShahzad: Woops, @antoviaque ^^

antoviaque:
@UmanShahzad 

> I imagined we'd do burst calls to the TransferWise API for recipient details using the recipient ID every month. We need to do it, right, if we're not storing that info? And even if we store it, if we need to update it we'd do that on TransferWise, still requiring API calls. I have not tested their API yet, but if we have a big enough team, I wonder if they have burst limits that'll cause trouble. Maybe I'm misunderstanding something in your vision here.

If we don't display the bank details on the invoice, then we probably don't need to retrieve these details at all. As mentioned on the other PR, we probably could get away with having the admin always going to transferwise to add that info, and then we just use the recipient id in our app - a bit like the way we would just handle card tokens with a stripe integration, where we never manipulate the actual card info.

UmanShahzad:
@antoviaque I see. In that case, the recipient ID would only be used to actually make transfers. If all of this meets our requirements, I'm happy to do it this way for sure! (Sorry I took a while to convince ;p ) Hopefully the road forward will be a little less bumpy as we figured this part out.

PR #7, opened 2018-01-05
WIP Vagrant Devstack

macornwell:


Vagrantfile
```diff
@@ -0,0 +1,15 @@
...
+
+  config.vm.network :forwarded_port, guest: 1786, host: 1786 
+  
+  config.ssh.forward_x11 = true
+  config.vm.provision 'shell',
+                      path: 'bin/setup',
+                      privileged: false,
+                      keep_color: true
+end
+
```
UmanShahzad: Nit: There's an extra newline here.

Vagrantfile
```diff
@@ -0,0 +1,15 @@
+Vagrant.configure("2") do |config|
+    config.vm.box = "ubuntu/trusty64"
```
UmanShahzad: Nit: match spacing throughout the file, either 2 or 4 but no mix.

bin/setup
```diff
@@ -0,0 +1,66 @@
+f this is a vagrant VM, cd to ~/opencraft on login
```
UmanShahzad: `sed -i 's/opencraft/accounting/' bin/setup`

UmanShahzad: Let's add `#!/bin/bash` at the top and a `set -e`.

Vagrantfile
```diff
@@ -0,0 +1,15 @@
...
+
+  config.vm.synced_folder '.', '/vagrant', disabled: true
+  config.vm.synced_folder '.', '/home/vagrant/opencraft'
```
UmanShahzad: `sed -i 's/opencraft/accounting/' Vagrantfile`

Vagrantfile
UmanShahzad: This is trusty, i.e. Ubuntu 14.04. We want [xenial, i.e. Ubuntu 16.04](https://app.vagrantup.com/ubuntu/boxes/xenial64). See [ubuntu version history](https://en.wikipedia.org/wiki/Ubuntu_version_history) as a reference.

bin/setup
UmanShahzad: "*If* this is a vagrant VM..."

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+if [[ $USER == 'vagrant' ]]; then
+    mkdir -p ~/opencraft
+    cd ~/opencraft
+    grep -Fq 'cd ~/opencraft' ~/.bashrc || echo 'cd ~/opencraft' >> ~/.bashrc
+fi
+
+# Need python-3.6 repository
+echo "Setting up python3.6 repository"
+sudo add-apt-repository ppa:jonathonf/python-3.6
```
UmanShahzad: Let's use [deadsnakes](https://github.com/open-craft/ansible-accounting/pull/1/files#diff-ce1279346898431a34ccea1a7ade5060R9), since I'm using that in live environments through Ansible, and if we use Ansible to provision this in the future, it'll be nice to be consistent.

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+sudo apt-get update
```
UmanShahzad: We might need a `-y` here?

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+
+sudo apt-get install -y python3.6 python3.6-dev
+
+# Install pip for python3
+echo "Installing pip for python3"
+sudo apt-get install -y python3-pip
+sudo pip3 install --upgrade pip
+sudo pip3 install --upgrade virtualenv
```
UmanShahzad: Let's make it one invocation?

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+sudo -E apt-get -y autoremove
+for i in `tr -d '\r' < debian_packages.lst`; do sudo apt-get install -y $i; done
+
+# Activate VirtualEnv
+cd ~/
+mkdir -p ~/.virtualenvs
+virtualenv -p python3.6 ~/.virtualenvs/opencraft
+source ~/.virtualenvs/opencraft/bin/activate
+
+# Need wkhtmlox
```
UmanShahzad: `sed -i 's/wkhtmlox/wkhtmltopdf/' bin/setup`. (WebKit HTML to PDF).

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+if [ ! -f "wkhtmlox-0.12.4_linux-generic-amd64.tar.xz" ]; then
+    echo "wkhtmltox not found. Downloading."
+    wnux-generic-amd64.tar.xzget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
```
UmanShahzad: What is `wnux-generic-amd64.tar.xzget`?

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+fi
+tar -xvf ./wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
+export PATH=$PATH:$HOME/wkhtmltox/bin
+
+
```
UmanShahzad: Nit: extra newline.

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+    grep -Fq 'source ~/.virtualenvs/opencraft/bin/activate' ~/.bashrc ||
+    echo 'source ~/.virtualenvs/opencraft/bin/activate' >> ~/.bashrc
+fi
+
+# Create postgres user
+sudo -u postgres createuser -d $USER ||
+  echo "Could not create postgres user '$USER' - it probably already exists"
+
+# Allow access to postgres from localhost without password
+cat << EOF | sudo tee /etc/postgresql/9.3/main/pg_hba.conf
```
UmanShahzad: We can't use 9.3; it has to be >= 9.4, since we use PostgreSQL's JSONField in Django, and that's a hard requirement for it. If you need, just install PostgreSQL through the official PPA. See https://github.com/open-craft/ansible-postgres/pull/3/files#diff-2444ad0870f91f17ca6c2a5e96b26823R1 where we do this.

bin/setup
```diff
@@ -0,0 +1,66 @@
...
+local   all             postgres                                peer
+local   all             all                                     trust
+host    all             all             127.0.0.1/32            trust
+host    all             all             ::1/128                 trust
```
UmanShahzad: Let's give the same HBA settings as enlisted [here](https://github.com/open-craft/ansible-playbooks/pull/4/files#diff-e2133f87124c722bd0cd073e929d8a6aR20), to keep it consistent.

debian_packages.lst
```diff
@@ -13,3 +13,6 @@ xvfb
 postgresql-client
 postgresql
 gettext
+libxml2-dev
+libxslt1-dev
+gettext
```
UmanShahzad: We're installing gettext on line 15. What are the others for, btw?

UmanShahzad:
@macornwell You'll have to rebase on `master` -- I just merged :) Also, I squashed, so I think you'll have to cherry-pick your commit on *top* of master, rather than rebasing.

bin/setup
macornwell: Good catch. the nux-gen... part was inserted between a wget.

bin/setup
macornwell: You seem confused. If you take a look at the download link at https://wkhtmltopdf.org/downloads.html you'll see that's the download for wkhtmltopdf.

macornwell: With the upgrade in ubuntu I can remove all this code now.

debian_packages.lst
macornwell: Removed and no longer needed with ubuntu upgrade. They were added to get python lxml running.

PR #8, opened 2018-01-09
Adding support for Vagrant.

macornwell:
This pull request creates support for Vagrant. In order to setup a vagrant devstack type 'vagrant up' which provisions the virtual machine.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,0 +1,60 @@
+#!/bin/bash 
+set -e
+
+# Fixes UTF-8 warning ("Your terminal is not configured to receive UTF-8 encoded text.") on some machines.
+export PYTHONIOENCODING="utf-8"
+grep -Fq 'export PYTHONIOENCODING' ~/.bashrc ||
+echo 'export PYTHONIOENCODING="utf-8"' >> ~/.bashrc
+
+# cd to ~/accounting on login
+if [[ $USER == 'ubuntu' ]]; then
+    mkdir -p ~/accounting
+    cd ~/accounting
+    grep -Fq 'cd ~/accounting' ~/.bashrc || echo 'cd ~/accounting' >> ~/.bashrc
+fi
+
+# Need python-3.6 repository
+echo "Setting up python3.6 repository"
+sudo add-apt-repository 'ppa:deadsnakes/ppa'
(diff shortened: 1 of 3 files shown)


bin/setup
```diff
@@ -0,0 +1,55 @@
...
+
+# Need python-3.6 repository
+echo "Setting up python3.6 repository"
+sudo add-apt-repository 'ppa:deadsnakes/ppa'
+sudo apt-get -y update
+
+# Install all dependencies.
+cd ~/accounting
+echo "Installing dependencies."
+sudo -E apt-get install -y `tr -d '\r' < debian_packages.lst`
```
UmanShahzad: Why doesn't this use the make target? Notice that it's now different, too, so this wouldn't work. Did you actually run this?

macornwell: I did not run it sense merging. After running it I've removed the comments. Why were comments added into the list file?

UmanShahzad: Xavier wanted me to explain in the file why we were downloading `wkhtmltopdf` when it's not actually being used through that package, to not confuse anyone.

macornwell:
@UmanShahzad  pushed up a fix. Good to go on not having time to fix it. I just ran it, everything works as expected.

bin/setup
```diff
@@ -31,6 +36,7 @@ if [[ $USER == 'ubuntu' ]]; then
     echo 'source ~/.virtualenvs/accounting/bin/activate' >> ~/.bashrc
 fi
 
+
```
itsjeyd: @macornwell Nit: Make whitespace between "sections" (delimited by inline comments) consistent -- there's two empty lines in some cases and a single empty line (which should probably be the default) in others.

bin/setup
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash 
 set -e
 
+# Fixes UTF-8 error message on some machines.
```
itsjeyd: @macornwell Nit: The UTF-8 thing is a warning, not an error. For a workaround like this it can be good to be explicit. How about:

```
# Fixes UTF-8 warning ("Your terminal is not configured to receive UTF-8 encoded text.") on some machines.
```

PR #9, opened 2018-01-09
Generate TransferWise Bulk Payment CSVs.

UmanShahzad:
This PR adds the ability to generate TransferWise bulk payment CSVs based off of the data we currently have available through the accounting service.

It currently behaves the same way generating invoice PDFs does: you create a model, enter a few pieces of data (in this case just the 'sender', which is OpenCraft), hit save with some checkboxes checked, and it'll generate it for all the appropriate users who're eligible for TransferWise, and upload the CSV to Google Drive automatically too.

**Testing Instructions**:

1. Setup the necessary users, accounts, bank accounts, hourly rates, etc., just as you would for invoices.
1. Make sure a TransferWise Recipient ID is set on the (only) bank account of those users who are connected to your OpenCraft client user through an hourly rate.
1. Go to http://localhost:1786/admin/transferwise/transferwisebulkpayment/add/.
1. Mark the `sender` as the OpenCraft client user and check all the marks.
1. See that a CSV file is in your media folder, as well as in the proper folder on GDrive (same as invoices).
1. See that the CSV file has columns that match https://transferwise.com/batch#/template-instructions?useSavedRecipients=true.
1. See that the CSV file's recipient information matches what you have in the local devstack.
1. (Optional) Add more valid recipients, generate their invoices, and make the bulk payment object, and see that all rows contain proper information.


UmanShahzad:
@itsjeyd Ready to go!

accounting/transferwise/models.py
```diff
@@ -0,0 +1,179 @@
...
+        if self.auto_create_csv_on_save:
+            self.to_bulk_payment_csv(self.payments.all())
+        if self.auto_upload_google_drive_on_save:
+            file = self.upload_to_google_drive(
+                file_path=self.csv_path,
+                target_path=[self.date.strftime('%Y'), 'invoices-in', self.date.strftime('%m')],
+                title=self.csv_filename,
+            )
+            self.csv_path = file['webContentLink']
+            TransferWiseBulkPayment.objects.filter(pk=self.pk).update(csv_path=self.csv_path)
```
itsjeyd: @UmanShahzad Same comment here about creating a separate method for the logic that updates `self.csv_path`.

accounting/transferwise/tests/test_mixins.py
```diff
@@ -0,0 +1,142 @@
...
+                'sourceCurrency',
+                'targetCurrency',
+                'amountCurrency',
+                'amount',
+                'paymentReference'
+            ])
+            with self.assertRaises(StopIteration):
+                next(reader)
+
+    def _create_payment(self, recipient_bank_account_kwargs=None):
```
itsjeyd: @UmanShahzad Nit: Move this above the test methods to make it easier to avoid mixing helper methods and test methods in the future.

accounting/transferwise/tests/test_models.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_auto_upload_google_drive_on_save(self, mock_upload_to_google_drive):
+        """ `upload_to_google_drive` is called when `auto_upload_google_drive_on_save` is checked. """
+        csv_path = 'https://drive.google.com/bulk_payment.csv'
+        mock_upload_to_google_drive.return_value = {'webContentLink': csv_path}
+        self.bulk_payment.auto_upload_google_drive_on_save = True
+        self.bulk_payment.save()
+        mock_upload_to_google_drive.assert_called_once()
+        self.assertEqual(self.bulk_payment.csv_path, csv_path)
+
+    def _create_recipients(self, invoices=False, invoice_date=NOW):
```
itsjeyd: @UmanShahzad Same nit here about moving this above the test methods.

itsjeyd: @UmanShahzad Also, let's call the second parameter `create_invoices` for clarity.

accounting/transferwise/tests/test_models.py
```diff
@@ -0,0 +1,153 @@
...
+class TransferWisePaymentTestCase(TestCase):
+    """ Test cases for `models.TransferWisePayment`. """
+
+    @freezegun.freeze_time(NOW)
+    def setUp(self):
+        """ Setup test objects. """
+        self.provider = AccountFactory(business_name='Developer')
+        self.client = AccountFactory(business_name='OpenCraft GmbH')
+        self.invoice = InvoiceFactory(provider=self.provider, client=self.client)
+        self.bulk_payment = factories.TransferWisePaymentFactory(date=NOW, invoice=self.invoice)
```
itsjeyd: @UmanShahzad This is for a single payment, so just call this `self.payment`?

accounting/common/logging.py
```diff
@@ -0,0 +1,52 @@
...
+    @staticmethod
+    def _namer(name):
+        """Append a .gz suffix to the rotated files."""
+        return name + '.gz'
+
+    @staticmethod
+    def _rotator(source, dest):
+        """Write compressed output when rotating files."""
+        with open(source, "rb") as sf:
+            with gzip.open(dest, "wb", compresslevel=9) as df:
```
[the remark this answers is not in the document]
UmanShahzad: This is copy and paste from https://github.com/open-craft/opencraft/blob/master/opencraft/logging.py. In order to ease maintenance between the two, I'd prefer to not touch this file. It might be a better long-term approach to move something like this out to a new, small library, uploaded to pip, but not sure.

accounting/account/models.py
```diff
@@ -89,6 +89,14 @@ def name(self):
         """
         return self.business_name or self.user.get_full_name() or self.user.username
 
+    @property
+    def active_bank_account(self):
+        """
+        Return the user's currently active bank account.
+        """
+        # TODO: Make it so only one bank account can be 'active' at a time.
```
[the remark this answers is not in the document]
UmanShahzad: If we associate a bank account with an invoice, that would read that an invoice belongs to a bank account, which doesn't sound like the correct relation to me. And both the client and provider on the invoice can have bank accounts.

If we associate it with an hourly rate, the latter problem again applies: a bank account for whom, the client or provider?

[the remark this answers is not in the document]
UmanShahzad: For a concrete future plan, I would think using something similar to edX's configuration model where you can set an object as 'active' would be the way to go. Or just change this relation to be 1-to-1 between bank account & user account.

accounting/common/tests/test_logging.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        # Initially, only the base log file exists
+        self.assertTrue(os.path.isfile(self.log_file))
+        for log_file in self.rotated_log_files:
+            self.assertFalse(os.path.isfile(log_file))
+
+        # As messages are logged, the rotated log files get created
+        for idx in range(self.num_logs + 1):
+            self.logger.error(self.test_string)
+            self.assertTrue(os.path.isfile(self.log_file))
```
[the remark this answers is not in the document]
UmanShahzad: See above.

accounting/invoice/models.py
```diff
@@ -332,7 +299,13 @@ def save(self, *args, **kwargs):
...
             self.to_pdf()
         if self.auto_upload_google_drive_on_save:
-            self.upload_to_google_drive()
+            file = self.upload_to_google_drive(
+                file_path=self.pdf_path,
+                target_path=[self.date.strftime('%Y'), 'invoices-in', self.date.strftime('%m')],
+                title=self.pdf_filename,
+            )
+            self.pdf_path = file['webContentLink']
+            Invoice.objects.filter(pk=self.pk).update(pdf_path=self.pdf_path)
```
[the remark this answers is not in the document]
UmanShahzad: Actually, it's not just to make it show up as a download link, but to properly set the variable, which is more important. If we don't do this, the variable won't be set because we already call the super's save above, and we can't call it again lest infinite recursion.

I will document this behavior.

accounting/transferwise/migrations/0001_initial.py
```diff
@@ -0,0 +1,64 @@
...
+                ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created')),
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
+                ('uuid', models.UUIDField(default=uuid.uuid4, editable=False, help_text='The universally unique identifier for this model instance.', verbose_name='UUID')),
+                ('date', models.DateTimeField(default=django.utils.timezone.now, help_text='The date this bulk payment was created and uploaded to TransferWise. Defaults to right now, but can be changed.')),
+                ('csv_path', models.URLField(blank=True, help_text='The absolute URL that can be used to retrieve the bulk payment CSV.', max_length=300, null=True)),
+                ('auto_create_payments_on_save', models.BooleanField(default=False, help_text='Whether TransferWise payments should automatically be created on save. Note that currently, this does not create actual TransferWise payments through TransferWise! It only creates them internally to track different data easily.')),
```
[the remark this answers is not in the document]
UmanShahzad: On a first read yeah, it's not too helpful, but it's also just a help text for display on the admin so it doesn't have to be too explicit for the administrator to know the behind-the-scenes in detail.

But it's basically referring to the fields that're actually on the model.

accounting/transferwise/models.py
```diff
@@ -0,0 +1,179 @@
...
+        for recipient in self.recipients:
+            try:
+                # Get an invoice matching this bulk payment's month & year.
+                invoice = recipient.provider_invoices.filter(
+                    client=self.sender,
+                    date__year=self.date.year,
+                    date__month=self.date.month
+                ).latest('date')
+            except Invoice.DoesNotExist:
+                # Fallback to the latest one, or just skip the recipient.
```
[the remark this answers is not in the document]
UmanShahzad: Ah yes, let's check for paid status :+1:

accounting/transferwise/models.py
UmanShahzad: Yep, see my comment above and let me know what you think about it.

accounting/transferwise/tests/test_mixins.py
UmanShahzad: :+1:

accounting/transferwise/tests/test_models.py
UmanShahzad: :+1:

UmanShahzad: :+1:

accounting/transferwise/models.py
```diff
@@ -0,0 +1,179 @@
...
+            sender=self.sender.user.username,
+            date=self.date.strftime("%Y-%m-%d"),
+        )
+
+    def create_payments(self):
+        """
+        Create the individual TransferWise payments needed for this bulk payment.
+
+        The chosen invoice is the one that matches this bulk payment's date's month for the given payment's
+        sender and recipient. If an invoice does not fit this criteria, it gets the latest one between the two.
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/transferwise/tests/test_models.py
UmanShahzad: Oh, that was a bad copy and paste. Jeez I'm bad with this.

accounting/common/logging.py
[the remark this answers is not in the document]
UmanShahzad: But we can certainly change it here *and* there if needed, however I didn't even read this code since I just knew what it did so I didn't bother fixing its readability.

UmanShahzad:
@itsjeyd Thanks for reviewing! The remaining test code is part of a separate ticket, yes. The test code is what took long, too, so I'm going to not increase the planned scope of this ticket & PR.

accounting/account/models.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad 

> ... Or just change this relation to be 1-to-1 between bank account & user account.

That seems like a good solution for now. In the future it would be cool for clients/providers to be able to select a bank account to use for a specific invoice. But unless we have another concrete use case that we need to satisfy now (?), I'd be in favor of simplifying things and moving to a 1-to-1 relation as you suggest.

accounting/common/logging.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad Ah, OK -- I wasn't aware that we had the same piece of code in OCIM. Definitely fine to keep this as is, then :)

accounting/common/tests/test_logging.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad 👍

accounting/transferwise/tests/test_models.py
itsjeyd: @UmanShahzad No worries :)

UmanShahzad:
@itsjeyd So the default value is already adjusted at https://github.com/open-craft/accounting/blob/uman/transferwise-bulk-payment/accounting/bank/migrations/0004_auto_20180104_1124.py#L21. The issue is that, yeah, old JSON values will have `accountNumber`. I'll write some specialized migration that changes `accountNumber` to `account_number` for all existing fields when the migration is run to use 0004.

accounting/account/models.py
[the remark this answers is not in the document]
UmanShahzad: Yeah, no use case to satisfy right now. But since we're already using this property to encapsulate the issue here, it would be more work to undo it, make it 1-to-1, and then in the future have to redo it by doing another migration from 1-to-1 to 1-to-many, etc.. So for the sake of time, and because this feature could be useful, I'd not remove it right now lest it lead to more unnecessary work.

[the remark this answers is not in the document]
itsjeyd: @UmanShahzad With all of the automation work we have coming up I have a feeling that this code will stay around for quite a while... Even though we might not implement a nice-to-have such as being able to select a bank account for a specific invoice in the near future.

But OK, let's wrap this up now -- we can see if Xavier has any feedback on this during upstream review, and make changes based on that (if necessary) :)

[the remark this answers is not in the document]
UmanShahzad: @itsjeyd Yeah that's a good point. I guess what I'm getting at is, at this point it's going to be more work to undo this and go to 1-to-1.

@antoviaque I think you'll probably want this to be 1-to-1 as well, but see my above comments: it's going to delay merging this PR and eat up time; it's working fine as is but yes the design could in principle be changed to match current use cases.

accounting/bank/migrations/0004_auto_20180104_1124.py
```diff
@@ -8,6 +8,30 @@
...
+    """
+    BankAccount = apps.get_model('bank', 'BankAccount')
+    for bank_account in BankAccount.objects.all():
+        # Update: `accountNumber` -> `account_number`.
+        if 'accountNumber' in bank_account.identification:
+            bank_account.identification['account_number'] = bank_account.identification['accountNumber']
+            del bank_account.identification['accountNumber']
+        # Add: `currency`.
+        if 'currency' not in bank_account.identification:
+            bank_account.identification['currency'] = ''
```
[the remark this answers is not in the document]
UmanShahzad: @itsjeyd Good catch: there exists a `DEFAULT_CURRENCY` setting, and we can make it default to this.

accounting/account/models.py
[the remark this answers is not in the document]
antoviaque: @itsjeyd @UmanShahzad Yup - I actually highly hesitated to point this out in the initial review, but then I thought I was already being quite a pain. Generally I'm in favor to keeping things simpler and closer to the existing use cases, rather than trying to plan ahead for things we might need one day. Usually that need never turns out to be needed, or not the way we initially thought, and we are just stuck with unnecessary complexity. So if it fits into the time you have allocated for addressing reviews this week, I would also be in favor of going towards the 1-to-1 relationship.

accounting/common/logging.py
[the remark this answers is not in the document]
antoviaque: @UmanShahzad Copy-pasting code is a pretty bright red flag - I don't think we should do this. Could you move that code to a repository containing shared utils code like this, which you then add to the requirements on both sides?

accounting/invoice/models.py
```diff
@@ -332,7 +299,14 @@ def save(self, *args, **kwargs):
         if self.auto_create_pdf_on_save:
             self.to_pdf()
         if self.auto_upload_google_drive_on_save:
-            self.upload_to_google_drive()
+            file = self.upload_to_google_drive(
+                file_path=self.pdf_path,
+                target_path=[self.date.strftime('%Y'), 'invoices-in', self.date.strftime('%m')],
+                title=self.pdf_filename,
+            )
+            # We can't save this field because we're already in the save method, so set it and then update it directly.
```
antoviaque: So should that need to happen *before* the `super().save()` method is called?

accounting/transferwise/admin.py
```diff
@@ -0,0 +1,42 @@
...
+from accounting.common.admin import UuidModelAdmin
+from accounting.transferwise import models
+
+
+@admin.register(models.TransferWiseBulkPayment)
+class TransferWiseBulkPaymentAdmin(UuidModelAdmin):
+    """ Admin configuration for the `TransferWiseBulkPayment` model. """
+    list_display = ('date', 'sender', 'csv_path',)
+    search_fields = ('sender__user__username',)
+    readonly_fields = ('csv_path',)
```
antoviaque: Could you make links like this clickable? https://stackoverflow.com/questions/1949248/how-to-add-clickable-links-to-a-field-in-django-admin seem

accounting/transferwise/models.py
```diff
@@ -0,0 +1,181 @@
...
+    A model that represents TransferWise bulk payments.
+
+    Can be used to generate bulk payment CSVs.
+    """
+
+    date = models.DateTimeField(
+        default=timezone.now,
+        help_text=_("The date this bulk payment was created and uploaded to TransferWise. "
+                    "Defaults to right now, but can be changed."))
+    sender = models.ForeignKey(
```
antoviaque: We don't have a use case where the sender is not OpenCraft, right? So I don't think we need this field, especially one that we have to select every time we create a bulk payment. Can this be replaced by a global config variable, or something similar that isn't set on every bulkpayment object?

accounting/transferwise/models.py
```diff
@@ -0,0 +1,181 @@
...
+        help_text=_("The sender of the TransferWise bulk payment. "
+                    "Refers to the account owner of the account initiating the payment."))
+    csv_path = models.URLField(
+        blank=True, null=True, max_length=300,
+        help_text=_("The absolute URL that can be used to retrieve the bulk payment CSV."))
+
+    # TODO: The following 3 fields are temporary.
+    # TODO: These will be automated through jobs at some point, generated at particular times of the month.
+    auto_create_payments_on_save = models.BooleanField(
+        default=False,
```
antoviaque: I'm not sure if I already asked about this in the previous PR, but why are the defaults to those options False, when we will always need them to be True in our use cases? That generates a bunch of clicks every time, so is there a good reason for this?

accounting/transferwise/models.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                # Get an invoice matching this bulk payment's month & year.
+                invoice = recipient.provider_invoices.filter(
+                    client=self.sender,
+                    date__year=self.date.year,
+                    date__month=self.date.month,
+                    paid=False,
+                ).latest('date')
+            except Invoice.DoesNotExist:
+                # Fallback to the latest unpaid one, or just skip the recipient.
```
antoviaque: Why do we do this? If we create a bulk payment for a given month, it doesn't sound correct to return invoices that are from a different month, even unpaid - that's counter-intuitive and will likely result in mistakes.

Also, what's the use case? A good one to consider could be the one we'll likely face next month with Kshitij, who might end up with both this month and next month's invoices processed in the next batch payment.

To handle this without needing to create two batch payments, could you use a start & end date for the bulk payment, with the start date set by default to an ancient date, and an end date set to now? Then include all unpaid invoices in the batch transfer, even if there are several of them - this way this would by default include all unpaid invoices to date in the transfer, but be explicit about it, and provide a way to override the behavior to select a specific time window.

accounting/transferwise/models.py
```diff
@@ -0,0 +1,181 @@
...
+                try:
+                    invoice = recipient.provider_invoices.filter(client=self.sender, paid=False).latest('date')
+                except Invoice.DoesNotExist:
+                    continue
+
+            try:
+                self.payments.create(bulk_payment=self, invoice=invoice)
+            except (IntegrityError, ValidationError):
+                continue
```
antoviaque: Why do we ignore these exceptions?

accounting/transferwise/models.py
```diff
@@ -0,0 +1,181 @@
...
+            self.create_payments()
+        if self.auto_create_csv_on_save:
+            self.to_bulk_payment_csv(self.payments.all())
+        if self.auto_upload_google_drive_on_save:
+            file = self.upload_to_google_drive(
+                file_path=self.csv_path,
+                target_path=[self.date.strftime('%Y'), 'invoices-in', self.date.strftime('%m')],
+                title=self.csv_filename,
+            )
+            # We can't save this field because we're already in the save method, so set it and then update it directly.
```
antoviaque: Same remark as above

accounting/common/logging.py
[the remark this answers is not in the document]
UmanShahzad: @antoviaque Yeah I think it'd be good to have a common repo for these kinds of things across our application repositories. That's going to be quite some work though so I'll make a ticket. :+1:

accounting/account/models.py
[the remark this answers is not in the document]
UmanShahzad: @antoviaque At this point I've gravitated towards that mindset too. Let's see if I have extra time available for this. Between having to write tests and this though, I will be prioritizing investing effort on tests.

accounting/invoice/admin.py
```diff
@@ -37,7 +37,7 @@ class InvoiceAdmin(UuidModelAdmin, SimpleHistoryAdmin):
 
 
 @admin.register(models.LineItem)
-class LineItemAdmin(UuidModelAdmin, SimpleHistoryAdmin):
+class LineItemAdmin(admin.ModelAdmin):
```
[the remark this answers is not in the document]
UmanShahzad: Because `LineItem` doesn't actually store a UUID and history, so this was bad copy and paste on my part.

accounting/invoice/models.py
UmanShahzad: I tried this and it fails miserably -- at this point I don't recall the error message. Either way I wouldn't bother with this: I'm deleting it in the very next PR that's up, #10.

accounting/transferwise/admin.py
UmanShahzad: No worries, I absolutely plan on doing this in https://tasks.opencraft.com/browse/OC-3637, for next week. Since it'll be a relatively simple PR, we can get it ready before the next billing cycle.

accounting/transferwise/models.py
UmanShahzad: @antoviaque If we do this the codebase will be frankenstein: in some places, like with invoices, we'll have generic clients, and here we'll have a pre-selected one. If we decide to undo this we should only do it if we undo the invoice stuff too, which is now going to be a lot of work since it's quite widespread.

But I think you don't need to worry about the manual work here: given my work in #10, I can easily create one extra job, to run on the 6th, that automatically makes this bulk payment CSV and assigns OpenCraft properly through the new `BILLING_CYCLE_USERS` setting, which is basically what you're talking about.

accounting/transferwise/models.py
UmanShahzad: @antoviaque I wrote that in from a dev perspective: it's easier to test it when you have a blank slate. But yeah, I get your point, from an admin perspective.

No worries, #10 is erasing these things. It won't be erasing the TransferWise ones just yet actually, but if I have time I will try to fit that in, and if not the new ticket, as mentioned above, is 1-2 points worth of work hopefully.

In the meantime, since Alicia will just be creating *1* of these, it's not too bad for 3-5 seconds, and will prevent me having to change up a bunch of test code for an hour.

accounting/transferwise/models.py
UmanShahzad: @antoviaque That's a good suggestion to use the start & end dates, yeah. I'll scratch out this invoice gathering code to just get all unpaid invoices in that range then. But we need to be sure, of course, that we're then marking invoices as paid properly. When do we consider an invoice paid, given the current state of the automation?

accounting/transferwise/models.py
UmanShahzad: @antoviaque This is mostly in place because you can hit 'save' and create these same payments twice. So it just covers that possible mistake. This won't be needed when we put this into an automated job though.

accounting/transferwise/models.py
UmanShahzad: Yep, same response as above in this case.

accounting/bank/migrations/0004_auto_20180104_1124.py
```diff
@@ -2,12 +2,36 @@
...
+    Delete: All old keys.
+    """
+    BankAccount = apps.get_model('bank', 'BankAccount')
+    for bank_account in BankAccount.objects.all():
+        # Update: `accountNumber` -> `account_number`.
+        if 'accountNumber' in bank_account.identification:
+            bank_account.identification['account_number'] = bank_account.identification['accountNumber']
+        # Add: `currency`.
+        if 'currency' not in bank_account.identification:
+            bank_account.identification['currency'] = settings.DEFAULT_CURRENCY
```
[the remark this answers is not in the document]
UmanShahzad: @antoviaque Ah, that's a good catch, when I was writing this I was thinking it was for the source currency, but yeah it's the target and wouldn't always be the correct default (and it's impossible to actually choose the correct default, except by querying TransferWise).

But yes this is because the CSV requires a target currency. The one that you generate from their UI already fills this part in for you, so that's presumably why you were seeing a locked-out target currency when trying to make a payment, since they just fetch it for you there too.

I just tested it out though, using their template, but making `targetCurrency` as `EUR` for everyone, and got this:

selection_003

So we should be good :+1:

UmanShahzad:
@antoviaque Alrighty, I addressed the review! The majority of your comments can be addressed in #10 because of how related the work there is to the solution to your comments, or in a follow-up ticket next week.

accounting/transferwise/models.py
antoviaque: @UmanShahzad I think you can consider an invoice paid once it has been included in a TransferWise bulk payment - for now that should be enough.

UmanShahzad: Done :+1:

accounting/account/models.py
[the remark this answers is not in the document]
UmanShahzad: Created https://tasks.opencraft.com/browse/OC-3702.

accounting/common/logging.py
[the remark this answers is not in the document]
UmanShahzad: Created https://tasks.opencraft.com/browse/OC-3703.

accounting/bank/migrations/0004_auto_20180104_1124.py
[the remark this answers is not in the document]
antoviaque: @UmanShahzad Thanks for checking this - and sounds good for the approach. CC @aliciaopencraft17

accounting/invoice/admin.py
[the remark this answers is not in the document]
antoviaque: @UmanShahzad OK - thanks for the explanation 👍

accounting/transferwise/admin.py
antoviaque: @UmanShahzad OK - could you add it explicitly to the scope of OC-3637 then, to not forget?

accounting/transferwise/models.py
antoviaque: @UmanShahzad This would be specific to the TransferWise payments. It can make sense to have different providers set for invoices, as you have both incoming and outgoing invoices. But we are only going to create TransferWise bulk payments for the payments we make. And even if it was a frankenstein, I would still prefer to not merge code like this, and instead log a ticket to undo it in the rest of the code base.

accounting/transferwise/models.py
antoviaque: @UmanShahzad Yup, in that case it's fine 👍

accounting/transferwise/models.py
UmanShahzad: @antoviaque I find it a little unsettling that we'd make a ticket to undo generic code to be un-generic -- that's the first I've heard of someone wanting that. This repository is open source code, so it's counterproductive to be doing this, considering you *won't have to set it yourself* given upcoming work.

accounting/transferwise/models.py
antoviaque: @UmanShahzad Could you add a TODO item there and an explanation of the scenario where this happens?

UmanShahzad: @antoviaque Sure. :+1:

accounting/transferwise/tests/test_models.py
```diff
@@ -79,18 +79,6 @@ def test_create_payments(self, date):
...
-        self._create_recipients(create_invoices=True)
-        for unused in range(3):
-            self.bulk_payment.create_payments()
-            self.assertEqual(models.TransferWisePayment.objects.all().count(), 2)
-
-    def test_create_payments_no_recipients(self):
-        """ Payments are not created because no valid recipients exist. """
-        self.bulk_payment.create_payments()
-        self.assertFalse(models.TransferWisePayment.objects.all().exists())
-
```
antoviaque: @UmanShahzad Could you test the behavior of the changed code in https://github.com/open-craft/accounting/pull/9/commits/e37fd90bd31be5ba82fca709adc5adf1c61a798d#diff-107f6db16b31f879de8edfef974b4f2eR119  rather than just delete tests?

UmanShahzad: The existing tests already do test the behavior, because factory-created invoices are unpaid. Do you mean to check if they got marked as paid successfully?

accounting/transferwise/models.py
antoviaque: @UmanShahzad I didn't ask to make the rest of the code ungeneric - only to remove the provider field in this specific model. And making things less generic can remove complexity - so yes, it can also be a good thing.

accounting/transferwise/admin.py
UmanShahzad: @antoviaque Done :+1:

accounting/transferwise/models.py
UmanShahzad: @antoviaque I understand you meant only this `sender` field. If we're going to delete this though, it'd honestly only make sense in a separate ticket that does it together with the `client` field on `Invoice`, to keep the codebase consistent in one go. Otherwise it won't reduce complexity, it'll just make it harder to understand how the pieces fit if reading from scratch. Even *I'd* be confused by it, as the person writing this stuff.

accounting/transferwise/tests/test_models.py
antoviaque: @UmanShahzad What I meant is more that since no test were added or modified after `start_date` and `end_date` were added, so it means that their behavior isn't tested no? Ie that it only includes invoices from the specified period? 

Also, are all the invoices used in these tests unpaid? It seem important to test that paid invoices are not included, even if they are in the time period entered?

UmanShahzad: @antoviaque Ah ok, that makes sense -- yeah I will test this behavior now.

accounting/transferwise/models.py
UmanShahzad: @antoviaque Made https://tasks.opencraft.com/browse/OC-3722 for this.

accounting/transferwise/tests/test_models.py
UmanShahzad: @antoviaque Covered in https://github.com/open-craft/accounting/pull/9/commits/817a7a372fb9959cdef4576fbf1445ddf971c53e

PR #10, opened 2018-01-16
Add automated jobs for the invoice approval flow.

UmanShahzad:
This PR adds monthly jobs that do the following:

* Sends invoice preparation notifications.
* Sends invoice approval requests.
* Sends finalized invoices.

**Testing Instructions**:

#### Endpoints

1. Make an invoice as usual. Get the invoice's UUID.
1. Go to `http://localhost:1786/invoice/{uuid}/approve/` and see that it gets marked as manually approved when it previously wasn't.
1. Go to `http://localhost:1786/invoice/{uuid}/approve/` again and see that you cannot approve an invoice when it's already approved.
1. Go to `http://localhost:1786/invoice/{uuid}/pay/` and see that it gets marked as paid.
1. Go to `http://localhost:1786/invoice/{uuid}/pay` again and see that you cannot pay an invoice when it's already paid.

#### Invoice Tasks

1. Delete the old invoice.
1. Make 3 accounts: 1 client, 2 providers, with an HourlyRate connection between the client and providers. Make sure the client's username matches what you have in `BILLING_CYCLE_USERS`. Make the client and providers have emails that're available to you, i.e. `me+client1@opencraft.com`, `me+provider1@opencraft.com`, `me+provider2@opencraft.com`, since you'll need to see the emails sent to providers with clients CC'd.

Testing invoice tasks locally and on stage will be slightly different:

##### Locally

1. Go into the shell with `make shell`, and run the following to make sure it comes out as expected:
1. `from accounting.invoice.tasks import *`
1. `send_invoice_prep_notification()`
1. You can temporarily disable GDrive uploads by switching the flag, so you can see the email message in your console rather than a huge wall of text representing the PDF, for the following two tasks:
1. `send_invoice_approval_request()`
1. `send_final_invoices()`

##### Stage

1. Go into the shell with `make shell`, and run the following to make sure it comes out as expected:
1. `from accounting.invoice.tasks import *`
1. For all of these, check that the email arrives for everyone with the proper subject, body, to, bcc, and cc, as appropriate:
1. `send_invoice_prep_notification()`
1. `send_invoice_approval_request()` -- check the `invoices-in/draft/{month}` folder for the invoice.
1. `send_final_invoices()` -- check the `invoices-in/{month}` folder for the invoice. If you did not approve the invoice manually by using the URL, check that it's marked as 'automatically' approved instead.

Preferably, do these twice, so you can see cases where the invoice doesn't exist when starting the pipeline, and cases where an invoice does exist.

## MERGE TODO

1. Make a `draft` folder on Accounting prod GDrive.


accounting/common/utils.py
```diff
@@ -18,25 +18,31 @@
 #
 
 """
-Management command for generating data (usually for a development environment) to play with.
+Invoice utilities.
```
UmanShahzad: This is the worst I've ever seen of git making a guess at a file move. What I really did here was delete the management command which was unfinished, and later on add this `utils.py` file. Git thinks I just changed names.

UmanShahzad:
@itsjeyd Ready for review & testing! This is now also on stage.

accounting/invoice/tests/test_views.py
```diff
@@ -20,3 +20,71 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, {'approved': True})
+        self.invoice.refresh_from_db()
+        self.assertTrue(self.invoice.is_approved)
+        self.assertEqual(self.invoice.approved, choices.InvoiceApproval.manually)
+
+        # Second visit when already approved doesn't change anything.
+        url = reverse('invoice:invoice-approve', [self.invoice.uuid])
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
itsjeyd: @UmanShahzad Hm... I'm not sure if 404 is the right status code to use here. After all, the resource still exists. Perhaps a 403 (Forbidden) that includes some info about why the server is refusing action would be more appropriate?

accounting/invoice/tests/test_views.py
```diff
@@ -20,3 +20,71 @@
...
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(response.data.get('paid'))
+        self.invoice.refresh_from_db()
+        self.assertTrue(self.invoice.paid)
+
+        # Second visit when already paid doesn't change anything.
+        url = reverse('invoice:invoice-pay', [self.invoice.uuid])
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
itsjeyd: @UmanShahzad Same comment here about the status code.

accounting/invoice/utils.py
```diff
@@ -21,27 +21,102 @@
...
-    """
-    Return the invoice number which defaults to `yyyy-mm` for the current year and month.
+    :param template: The email template.
+    :param subject: The email subject.
+    :param extra_email_context: Extra kwargs to update the email context with.
+    :param client_usernames: The list of billing cycle users' usernames.
+    :param create_invoice: Whether to create a new invoice based off of a past one.
+    :param draft_invoice: Whether this is a draft invoice or not.
+    :param fill_line_items_from_jira: Whether to fill up the invoice's line items from JIRA worklogs.
+    :param upload_to_google_drive: Whether to upload the resulting invoice PDF to Google Drive.
```
itsjeyd: @UmanShahzad Could you extend this to describe the purpose of "draft invoices", where this code will store them, etc.?

accounting/invoice/utils.py
```diff
@@ -21,27 +21,102 @@
...
     """
-    return timezone.now().strftime("%Y-%m")
+    if extra_email_context is None:
+        extra_email_context = {}
+    now = datetime.datetime.now()
+    past = (now - datetime.timedelta(days=now.day))
+    past_month_formatted = past.strftime('%B')
```
itsjeyd: @UmanShahzad Logic for computing (and formatting) the past month appears in multiple places in this diff -- worth putting into a small utility function?

accounting/invoice/utils.py
```diff
@@ -21,27 +21,102 @@
...
-    Return the `datetime` object representing the first day of the month that just passed.
+# pylint: disable=too-many-arguments,too-many-locals
+def send_email_with_invoice(template, subject, extra_email_context=None, client_usernames=settings.BILLING_CYCLE_USERS,
+                            create_invoice=False, draft_invoice=False, fill_line_items_from_jira=False,
+                            upload_to_google_drive=False, auto_approve=False):
     """
-    return get_last_day_past_month().replace(day=1)
+    Send an email with an invoice attached, and potentially perform peripheral actions.
 
+    TODO: This function's pretty beefy; break it up by encapsulating logic into more utility functions or a new class.
```
itsjeyd: @UmanShahzad Are you planning on doing this in the context of the current task (https://tasks.opencraft.com/browse/OC-3669)? If none of the tasks that are scheduled for the next couple sprints are going to involve touching this code, it would probably be best to do this refactoring now.

accounting/invoice/utils.py
```diff
@@ -21,27 +21,102 @@
...
+                    provider=invoice.provider,
+                    client=invoice.client,
+                    extra_text=invoice.extra_text,
+                    extra_image=invoice.extra_image,
+                    template=invoice.template,
+                )
+            if fill_line_items_from_jira:
+                invoice.fill_line_items_from_jira()
+
+            # Make the email, attach the invoice, and and send it.
```
itsjeyd: @UmanShahzad Typo: "and and"

accounting/invoice/utils.py
```diff
@@ -21,27 +21,102 @@
...
 
+from accounting.account.models import Account
+from accounting.invoice import choices, models
 
-def get_last_day_past_month():
+
+def resolve_dict_callables(dictionary, *args):
     """
-    Return the `datetime` object representing the last day of the month that just passed.
+    Resolve any potential callables in a dictionary by passing it some arguments.
```
itsjeyd: @UmanShahzad This would benefit from a concrete example :) You could tailor the example to the use case below (where `message_context` is updated with additional data from the result of doing `resolve_dict_callables(extra_email_context, invoice)`).

accounting/invoice/views.py
```diff
@@ -20,3 +20,47 @@
...
+    """
+    A view for retrieving and updating `Invoice` objects.
+    """
+
+    permission_classes = permissions.IsAdminUser
+    queryset = models.Invoice.objects.all()
+    lookup_field = 'uuid'
+
+    @decorators.detail_route(
+        permission_classes=[permissions.AllowAny],
```
itsjeyd: @UmanShahzad Just to confirm, this endpoint is still protected by the class' `permission_classes` (permission.IsAdminUser), right?

UmanShahzad: Only these are protected by `IsAdminUser`:

1. /invoice/
1. /invoice/{uuid}
1. /invoice/{uuid}/pay/

We need `/invoice/{uuid}/approve/` to be open since otherwise no one could confirm their invoice.

accounting/invoice/utils.py
UmanShahzad: It's likely that I'll be touching this code for https://tasks.opencraft.com/browse/OC-3697, so I put this TODO here for that.

accounting/invoice/utils.py
UmanShahzad: Yes, sure :+1:

accounting/invoice/utils.py
UmanShahzad: Yes, true, I should probably do that :+1:

accounting/invoice/utils.py
UmanShahzad: :+1:

accounting/invoice/utils.py
UmanShahzad: Sure, I can give it an example :+1:

accounting/invoice/tests/test_views.py
UmanShahzad: Actually, it makes sense given how the view handles this: when trying to `approve` or `pay` an invoice, you can only do that for invoices that aren't already approved or paid, respectively. So the view only filters invoices with the appropriate condition, i.e. not approved or not paid. So if you get a 404 it means it could not find an invoice for approval or an invoice that's still payable. Of course if you remove the verb, it will find it (if you have admin access).

accounting/invoice/tests/test_tasks.py
```diff
@@ -0,0 +1,190 @@
...
+            "\n"
+            "I have attached your finalized and approved invoice for December.\n"
+            "\n"
+            "We are now proceeding to send the bank transfers, and you can expect payment to arrive "
+            "in your bank account within 5-8 business days at most.\n"
+            "\n"
+            "If you have any issues, please contact billing@opencraft.com.\n"
+        )
+        self.assertEqual(mail.outbox[0].body, body.format(self.invoice1.uuid))
+        self.assertEqual(mail.outbox[1].body, body.format(self.invoice2.uuid))
```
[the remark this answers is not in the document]
UmanShahzad: Yeap I'm just trying to get this out the way given the time I have for such a large feature surface. I could always break tasks down and have just done a single function in `invoice.tasks` per week, allowing myself plenty of time to refactor this quite well, but that wouldn't suffice our requirements.

accounting/invoice/tests/test_tasks.py
```diff
@@ -0,0 +1,190 @@
...
+        # Body contains proper invoice approval URL and text, for both providers.
+        body = (
+            "Hello,\n"
+            "\n"
+            "I have prepared your invoice for December! It is attached and ready for you to review.\n"
+            "\n"
+            "When you've looked it over, can you approve it by clicking on this link, "
+            "or by visiting it in your browser?\n"
+            "\n"
+            "https://billing.opencraft.com/invoice/{}/approve/\n"
```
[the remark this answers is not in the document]
UmanShahzad: Sure, but I wouldn't think it's really necessarily since we already format with `.uuid`.

accounting/invoice/tests/test_tasks.py
```diff
@@ -0,0 +1,190 @@
...
+            "\n"
+            "If you find any issues with your invoice, please contact billing@opencraft.com before January 5th -- "
+            "I will consider your invoice approved by then, and will proceed to sending the corresponding payment.\n"
+        )
+        invoices = models.Invoice.objects.order_by('date')[2:]
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/invoice/tests/test_models.py
```diff
@@ -20,3 +20,61 @@
...
+@ddt.ddt
+class LineItemTestCase(TestCase):
+    """ Test cases for `models.LineItem`. """
+
+    def setUp(self):
+        """ Set up test objects. """
+        self.line_item = factories.LineItemFactory(
+            line_item_id=1,
+            key='OC-4000',
+            name='Accounting: Refactor bad test code from OC-3589',
```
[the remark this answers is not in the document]
UmanShahzad: :)

accounting/invoice/tests/test_choices.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# OpenCraft -- tools to aid developing and hosting free software projects
```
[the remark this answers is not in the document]
UmanShahzad: Nope! I'm still waiting for @antoviaque to let me know when a consensus is reached.

accounting/invoice/migrations/0006_auto_20180116_1745.py
```diff
@@ -0,0 +1,49 @@
...
+        ),
+        migrations.AddField(
+            model_name='historicalinvoice',
+            name='approved',
+            field=models.CharField(choices=[('automatically', 'Automatically'), ('manually', 'Manually'), ('not_approved', 'Not Approved')], default='not_approved', help_text='The approval status of this invoice.', max_length=80),
+        ),
+        migrations.AddField(
+            model_name='invoice',
+            name='approved',
+            field=models.CharField(choices=[('automatically', 'Automatically'), ('manually', 'Manually'), ('not_approved', 'Not Approved')], default='not_approved', help_text='The approval status of this invoice.', max_length=80),
```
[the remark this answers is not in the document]
UmanShahzad: I had gone with a BooleanField before, but @antoviaque wanted to be able to track whether it was approved manually or automatically. I then kept the names succinct because reading `invoice.approved == automatically` feels more right than `invoice.approved = automatically_approved`, which reads with some redundancy IMO.

Let's stick with this for now? I'd have to change I think a lot of places in the code for this and fix up migrations on stage.

.environ/.env.dev
```diff
@@ -32,6 +32,23 @@ ENABLE_GOOGLE=true
 GOOGLE_AUTH_CLIENT_SERVICE_EMAIL='bot-name@project-name.iam.gserviceaccount.com'
 GOOGLE_DRIVE_ROOT='root-folder-id'
 
+# BILLING CYCLE ###############################################################
+
+# List of users to consider in the billing cycle whose providers are sent invoice-related emails
+# every month at specific dates. See `accounting.invoice.tasks`.
+BILLING_CYCLE_USERS='opencraft'
```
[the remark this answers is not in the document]
UmanShahzad: Yes sure, although this is more of a Linux thing: you generally separate these env variables with commas. I would actually upstream this documentation to `environ` which handles this. But little time left this week, so will just leave a comment here.

accounting/bank/tests/test_models.py
```diff
@@ -20,3 +20,88 @@
...
+            (
+                ('BIC/SWIFT', 'asdf'),
+                ('Account Number', 'asdf2'),
+                ('Currency', 'asdf3'),
+            ),
+        ),
+    )
+    @ddt.unpack
+    def test_existing_identification(self, identification, expected_output):
+        """ The `existing_identification` function returns only what's available. """
```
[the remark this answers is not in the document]
UmanShahzad: Haha, well I thought it would be clear why `not_a_real_choice` is not a real choice ;)

accounting/invoice/tasks.py
```diff
@@ -0,0 +1,94 @@
...
+    past_month = (now - datetime.timedelta(days=now.day)).strftime('%B')
+    template = get_template(constants.INVOICE_NOTIFICATION_TEMPLATE)
+    for client_username in settings.BILLING_CYCLE_USERS:
+        client = Account.objects.get(user__username=client_username)
+        provider_emails = [rate.provider.user.email for rate in client.client_hourly_rates.all()]
+        message = template.render({'approval_date': approval_date, 'month': past_month})
+        EmailMessage(
+            subject=constants.INVOICE_NOTIFICATION_SUBJECT.format(month=past_month),
+            body=message,
+            bcc=provider_emails,
```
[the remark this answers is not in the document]
UmanShahzad: Well actually in this case, there's no invoice being attached, because this is just for the original notification to prepare for the invoice on the 3rd. Sending a single email with an attached file that should differ per recipient isn't possible.

Django does have a `send_mass_email` function, which opens a single connection to send emails to multiple `to` targets, but it ends up including all of their names in the list, which is something I wanted to hide, with `bcc`.

accounting/invoice/tests/test_tasks.py
```diff
@@ -0,0 +1,190 @@
...
+    def setUp(self):
+        """ Set up test objects. """
+        self.create_client_and_provider_links()
+        super().setUp()
+
+    def test_send_invoice_prep_notification(self):
+        """ The invoice preparation notification email sent out every month contains proper information. """
+        tasks.send_invoice_prep_notification()
+        # Only a single email is sent out per client, with all providers in the BCC list.
+        self.assertEqual(len(mail.outbox), 1)
```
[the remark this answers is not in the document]
UmanShahzad: Good point :+1:

accounting/invoice/tests/test_views.py
UmanShahzad: Yep, same reply for this case.

accounting/invoice/views.py
```diff
@@ -48,10 +48,7 @@ def approve(self, request, uuid=None):
         invoice.save()
         return response.Response({'approved': True})
 
-    @decorators.detail_route(
-        permission_classes=[permissions.IsAdminUser],
```
UmanShahzad: Realized it's redundant since it uses the parent permission.

accounting/invoice/views.py
itsjeyd: @UmanShahzad Ack, of course -- for some reason I thought I was looking at the `pay` endpoint. Sorry for the noise!

accounting/invoice/migrations/0006_auto_20180116_1745.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad Hm, I agree that `invoice.approved == automatically` reads well... but personally don't mind `invoice.approved = automatically_approved` too much.

The main thing for me is that it feels weird to mix "automatically"/"manually"/"not approved" -- it's like the field can't decide whether it wants to be a Boolean or not ;) "automatically approved"/"manually approved"/"not approved" seems to make a tad more sense as a triple of acceptable values.

I'm fine with covering this as part of a separate ticket though. We can also let @antoviaque weigh in first (i.e., before creating a ticket or adding this item to an upcoming ticket that already exists), to see if he has a strong preference one way or another.

accounting/invoice/tasks.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad Man... I'm a little tired today! What I meant to say was:

> This makes it possible to send a single e-mail and still have each provider receive only their own invoice **notification**

Gah! :)

accounting/invoice/tests/test_models.py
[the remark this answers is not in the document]
itsjeyd: It better be something good! And accounting-related... 😛

accounting/invoice/tests/test_tasks.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad I actually think the existing code reads really well. It can be nice and/or helpful for test code to be a bit more verbose. And sometimes figuring out where to draw the line isn't straightforward. So just in case my previous comment came across as though I was complaining: I wasn't :)

accounting/invoice/utils.py
```diff
@@ -21,27 +21,115 @@
...
+    for client_username in client_usernames:
+        client = Account.objects.get(user__username=client_username)
+        providers = [rate.provider for rate in client.client_hourly_rates.all()]
+        for provider in providers:
+            # Copy over certain details from the latest invoice from the proper month to the new invoice.
+            invoice = models.Invoice.objects.filter(
+                provider=provider,
+                client=client,
+                date__month=past.month if draft_invoice else now.month,
+            ).latest('date')
```
itsjeyd: @UmanShahzad When running `send_invoice_approval_request` this fails with `DoesNotExist: Invoice matching query does not exist.` if there is no invoice for the previous month, which is an edge case that we'll need to cover. (Everybody's getting paid for the first time at some point :))

UmanShahzad: @itsjeyd @antoviaque Yep, I ran into this already, but the issue is that we need some kind of 'template' of the invoice because it has to copy over `extra_text`, `extra_image`, etc., which would *have* to be copied from somewhere.

This would require a significant refactor, to use a new `InvoiceTemplate` class for each provider that contains all of these fields, which would have a 1-to-many relationship with `Invoice`s themselves, so that they can pull information from the template. This way for new team members we'd just need to create the template and then the invoice will use that for its per-provider info.

Since that's going to be a lot of changes (mostly names and custom migration code) and I only thought about it today, then for all new team members who use our automation, Alicia will have to create an initial dummy paid invoice with the template details. This dummy invoice, if created for the previous month and already marked paid, won't have an effect on our transferwise bulk payments, and won't have any line items associated with them. When we move to using real templates we can delete the dummy invoices.

I can work on this next week, but it's the last week before the billing cycle, so no guarantee that it'll be deployed to prod by the 1st.

accounting/invoice/views.py
```diff
@@ -20,3 +20,44 @@
...
+        """
+        Approve the invoice with UUID `uuid`.
+        """
+        invoice = self.get_object()
+        invoice.approved = choices.InvoiceApproval.manually
+        invoice.save()
+        return response.Response({'approved': True})
+
+    @decorators.detail_route(queryset=models.Invoice.objects.filter(paid=False))
+    def pay(self, request, uuid=None):
```
antoviaque: @UmanShahzad Where is the `pay` view linked from? The invoices are marked as paid from the admin, no? If so why do we need this view?

accounting/invoice/utils.py
antoviaque: @UmanShahzad Rather than a whole separate model, these could be fields on the provider object for now? Ie `provider.default invoice_extra_text`, etc. - this way at least the current PR would have a proper behavior, without weirdness like these that we'll need to remember, and without adding too much refactoring work. Then the InvoiceTemplate could be implemented in the future if/when we need to add more similar fields or custom templating behavior.

accounting/invoice/utils.py
```diff
@@ -21,27 +21,115 @@
...
+            if create_invoice:
+                invoice = models.Invoice.objects.create(
+                    number=invoice.number,
```
antoviaque: @UmanShahzad Shouldn't that number be incremented? The default format is YYYY-MM if I remember correctly? Or YYYY-XX, with XX starting from 01?

accounting/invoice/choices.py
```diff
@@ -37,3 +37,22 @@ class InvoiceTemplate(DjangoChoices):
     By default, the `default/default.html` template is used, with styles `default/css/default.css`.
     """
     Default = ChoiceItem('default', _('Default'))
+
+
+class InvoiceApproval(DjangoChoices):
```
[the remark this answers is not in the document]
UmanShahzad: :+1:

accounting/invoice/templates/emails/invoice_notification.txt
```diff
@@ -0,0 +1,7 @@
+Hello,
+
+At the beginning of each month, to help you gain time in invoicing your hours, I automatically build an invoice based on the worklogs you logged in Jira the month before.
+
+I just wanted to let you know that I'll be doing this soon -- I'll send you an invoice for {{ month }} to approve on {{ approval_date|date:"F jS" }}.
+
+Stay tuned!
```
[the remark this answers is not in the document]
UmanShahzad: Sure :+1:

accounting/invoice/views.py
UmanShahzad: I realized when writing the view above it that a similar `pay` one would be really fast to do, and is going to be necessary when the UI comes out. So I just did it. Let's not erase that, it's not going to bite anyone.

accounting/invoice/utils.py
UmanShahzad: Hmm, that's true, but that actually presents a new problem I didn't think about. People have totally different schemes for this: some with the default format, some with OC-XX, some with just XX, and so on. There are ways to handle this, like just iterate through a list of possible schemes and see if there's a match, and if so use that scheme's iterating method on the number. I can go through the current schemes that people in the team are using and build this, but it would have to be maintained for new schemes.

accounting/invoice/tasks.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Send emails to notify invoice recipients to prepare for invoice receipt.
+    """
+    now = datetime.datetime.now()
+    approval_date = now.replace(day=int(settings.INVOICE_APPROVAL_DAY))
+    past_month = get_last_day_past_month().strftime('%B')
+    template = get_template(constants.INVOICE_NOTIFICATION_TEMPLATE)
+    for client_username in settings.BILLING_CYCLE_USERS:
+        client = Account.objects.get(user__username=client_username)
+        provider_emails = [rate.provider.user.email for rate in client.client_hourly_rates.all()]
```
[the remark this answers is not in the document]
UmanShahzad: @antoviaque We can just add an `active` field on the account, for active/deactivated accounts.

accounting/invoice/utils.py
UmanShahzad: @antoviaque Hmm.. instead, I'll just make `InvoiceTemplate`, but skip any migration work. It's equivalent to what you're suggesting, except it builds the initial code infrastructure for the model, which is a good step forward, and that invoices would not be linked to it: each `Account` (i.e. provider) will be linked to it. Then in the future we can remove that relation in place of 1-to-many with InvoiceTemplate-to-Invoice.

accounting/invoice/tests/test_tasks.py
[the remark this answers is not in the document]
UmanShahzad: Haha, yep no worries :+1:

accounting/invoice/utils.py
UmanShahzad: I bit the bullet and went ahead and did this -- luckily it's not a mess, and in fact very readable code. Will push later.

accounting/invoice/utils.py
UmanShahzad: I went the mile: implementing the whole thing right now properly.

accounting/invoice/utils.py
antoviaque: @UmanShahzad Just seeing this comment now - it would be fine to have a simple approach on this for now, ie the most common use-case(s), and do the rest by hand. No need to solve all the problems at once : ) Good that you found a simple solution already though!

UmanShahzad:
@antoviaque @itsjeyd I addressed your reviews, and also went the mile and implemented some of the features we thought would be difficult. :) With the help of the InvoiceTemplate, things are looking a lot better.

accounting/invoice/tasks.py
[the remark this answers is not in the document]
UmanShahzad: All good :+1:

accounting/invoice/views.py
UmanShahzad: :+1:

accounting/invoice/utils.py
UmanShahzad: @antoviaque Yep, I just went ahead and did it, and it'll prevent any further manual work from here on. (One would have to do the manual work no matter what every month otherwise).

accounting/account/models.py
```diff
@@ -117,6 +117,9 @@ class HourlyRate(CommonModel, UuidModel):
     client = models.ForeignKey(
         Account, models.CASCADE, related_name='client_hourly_rates', blank=True, null=True,
         help_text=_("The account linked to the user who pays the hourly rate to the provider."))
+    active = models.BooleanField(
```
UmanShahzad: This field allows us to more easily deal with tracking whether someone is still under contract or not. I tried doing that through checking if JIRA worklogs existed, but it complicated things a lot more and gave rise to new edge cases, so I dropped it in favor of this.

itsjeyd:
Thanks @UmanShahzad, I'll have a look at this now.

accounting/invoice/choices.py
```diff
@@ -56,3 +60,126 @@ def approved_choices(cls):
...
+        Increment any passed in value given its matching numbering scheme choice.
+
+        Assumes, if given:
+        * `choice` is one of the value strings associated with a choice, not the `ChoiceItem` object
+        * `value` is in the format of the choice's numbering scheme.
+        """
+        function_name = cls.attributes[choice]
+        function = getattr(InvoiceNumberingScheme, 'increment_{}'.format(function_name))
+        if value is None:
+            value = getattr(InvoiceNumberingScheme, 'default_{}'.format(function_name))()
```
itsjeyd: @UmanShahzad Under what conditions will `value` be `None`? It could be worth mentioning that here. If it'll be `None` if and only if we're generating an invoice for a team member for the first time, shouldn't this method return the default value? The way it's currently written it seems like it would increment the default value before returning it.

accounting/invoice/migrations/0008_invoice_templates.py
```diff
@@ -0,0 +1,128 @@
...
+                ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created')),
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
+                ('uuid', models.UUIDField(default=uuid.uuid4, editable=False, help_text='The universally unique identifier for this model instance.', verbose_name='UUID')),
+                ('numbering_scheme', models.CharField(choices=[('%Y-%m', 'yyyy-mm (2018-01)'), ('{number}', '{number} (42, for the 42nd invoice)'), ('%Y-%m-{number}', 'yyyy-mm-{number} (2018-01-42, for the 42nd invoice)'), ('OC-%Y-%m', 'OC-yyyy-mm (OC-2018-01)'), ('OC-{number}', 'OC-{number} (OC-42, for the 42nd invoice)')], default='%Y-%m', help_text='The numbering scheme used to determine how to increment the invoice number.', max_length=80)),
+                ('extra_text', models.TextField(blank=True, help_text='Any arbitrary extra text that the provider would like to display on their invoice. Each template should have a designated location to place this extra text.', null=True)),
```
itsjeyd: @UmanShahzad 

> Each *HTML* template should have a designated location to place this extra text.

Or even:

> *The HTML template that belongs to this invoice template* should have a designated location to place this extra text.

accounting/invoice/tasks.py
```diff
@@ -49,7 +49,7 @@ def send_invoice_prep_notification():
     template = get_template(constants.INVOICE_NOTIFICATION_TEMPLATE)
     for client_username in settings.BILLING_CYCLE_USERS:
         client = Account.objects.get(user__username=client_username)
-        provider_emails = [rate.provider.user.email for rate in client.client_hourly_rates.all()]
+        provider_emails = [rate.provider.user.email for rate in client.client_hourly_rates.all() if rate.active]
```
itsjeyd: @UmanShahzad Instead of fetching `all` rates from the DB, this could use `client.client_hourly_rates.filter(active=True)`.

accounting/invoice/tests/test_choices.py
```diff
@@ -35,3 +43,97 @@ def test_approved_choices(self):
...
+        ('000003', '4'),
+    )
+    @ddt.unpack
+    def test_increment_number(self, value, incremented_value):
+        """ Incrementing the `number` choice gives the correct number, including with leading 0s. """
+        self.assertEqual(choices.InvoiceNumberingScheme.increment_number(value), incremented_value)
+
+    @ddt.data(
+        ('2018-01-4', '2018-02-5'),
+        ('2018-12-10', '2019-01-11'),
```
itsjeyd: @UmanShahzad I'm assuming that for the year-month-number scheme, we might get values like `2018-05-37` or even `2018-05-257`, right? Maybe we can make that more visible by including such a value in the data defined here.

accounting/invoice/tests/test_choices.py
```diff
@@ -35,3 +43,97 @@ def test_approved_choices(self):
...
+        """ Incrementing the `opencraft_number` choice gives the correct prefix and number including for leading 0s. """
+        self.assertEqual(choices.InvoiceNumberingScheme.increment_opencraft_number(value), incremented_value)
+
+    @ddt.data(
+        (choices.InvoiceNumberingScheme.default, None, '2018-02'),
+        (choices.InvoiceNumberingScheme.default, '2018-12', '2019-01'),
+        (choices.InvoiceNumberingScheme.number, '2', '3'),
+        (choices.InvoiceNumberingScheme.year_month_number, '2018-12-3', '2019-01-4'),
+        (choices.InvoiceNumberingScheme.opencraft_year_month, 'OC-2018-09', 'OC-2018-10'),
+        (choices.InvoiceNumberingScheme.opencraft_number, 'OC-0002', 'OC-3'),
```
itsjeyd: @UmanShahzad If someone wants a certain number of leading zeros (or a certain number of digits in the number part of this format), shouldn't we preserve that?

accounting/invoice/utils.py
```diff
@@ -85,22 +106,27 @@ def send_email_with_invoice(template, subject, extra_email_context=None, client_
     template = get_template(template)
     for client_username in client_usernames:
         client = Account.objects.get(user__username=client_username)
-        providers = [rate.provider for rate in client.client_hourly_rates.all()]
+        providers = [rate.provider for rate in client.client_hourly_rates.all() if rate.active]
```
itsjeyd: @UmanShahzad Same comment here about using `filter` instead of `all`.

accounting/invoice/utils.py
```diff
@@ -85,22 +106,27 @@ def send_email_with_invoice(template, subject, extra_email_context=None, client_
...
+                invoice = models.Invoice.objects.filter(
+                    provider=provider,
+                    client=client,
+                    date__month=past.month if draft_invoice else now.month,
+                ).latest('date')
+            except models.Invoice.DoesNotExist:
+                # This may be the provider's first invoice.
+                invoice = None
+
+            if create_invoice or not invoice:
```
itsjeyd: @UmanShahzad Would it make sense to remove the `create_invoice` parameter from this function and default to always creating an invoice based on the latest data? That way the admin wouldn't need to remember to do this, and we'd make sure that the system wouldn't send a stale invoice.

accounting/invoice/utils.py
```diff
@@ -85,22 +106,27 @@ def send_email_with_invoice(template, subject, extra_email_context=None, client_
...
-                    number=invoice.number,
-                    provider=invoice.provider,
-                    client=invoice.client,
-                    extra_text=invoice.extra_text,
-                    extra_image=invoice.extra_image,
-                    template=invoice.template,
+                    number=choices.InvoiceNumberingScheme.increment_value(
+                        invoice.template.numbering_scheme,
+                        value=invoice.number
+                    ) if invoice else choices.InvoiceNumberingScheme.default_default(),
```
itsjeyd: @UmanShahzad We can also get at the `numbering_scheme` to use via `provider.invoice_template.numbering_scheme`, so it'd be good to use that here instead of falling all the way back to `default_default`.

accounting/invoice/choices.py
```diff
@@ -56,3 +60,126 @@ def approved_choices(cls):
...
+    def increment_default(cls, value):
+        """
+        Increment the number for the 'default' numbering scheme.
+
+        Assumes `value` is in the format of the 'default' numbering scheme.
+        """
+        # Turn the input value string into a datetime object.
+        datetime_obj = datetime.datetime.strptime(value, cls.default)
+        # Increment it one month forward.
+        incremented_obj = datetime_obj + datetime.timedelta(days=32)
```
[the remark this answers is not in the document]
UmanShahzad: @itsjeyd Hmm, nope, I think that would be good too.

accounting/invoice/choices.py
UmanShahzad: @itsjeyd Oh damn, yeah, I mixed up the reason I even implemented it this way: in the `send_email_with_invoice` function, when we're generating for the first time, we can pass in `None` to some `default_value` function, not an *increment* function. I will change this and its use case.

accounting/invoice/migrations/0008_invoice_templates.py
UmanShahzad: :+1:

accounting/invoice/tasks.py
UmanShahzad: @itsjeyd Argh, good point. :p

accounting/invoice/tests/test_choices.py
UmanShahzad: :+1:

accounting/invoice/tests/test_choices.py
UmanShahzad: @itsjeyd I would want to but this adds complexity to incrementing the values. I think only @tomaszgy has it this way -- would this be OK with you, @tomaszgy ?

accounting/invoice/utils.py
UmanShahzad: :+1:

accounting/invoice/utils.py
UmanShahzad: @itsjeyd But then this creates multiple invoices since we'd go through this twice (once for the approval, and once for finalization). This then messes with how we decided to use TransferWiseBulkPayments -- get all unpaid invoices for a certain date range (from ancient history to now by default).

accounting/invoice/utils.py
UmanShahzad: @itsjeyd Oh yes, good point.

accounting/invoice/tests/test_choices.py
tomaszgy: @UmanShahzad Please excuse but I'm not sure what I'm being asked here :) Could you please say a bit more?

accounting/invoice/utils.py
```diff
@@ -122,8 +122,8 @@ def send_email_with_invoice(template, subject, extra_email_context=None, client_
                 invoice = models.Invoice.objects.create(
                     number=choices.InvoiceNumberingScheme.increment_value(
                         invoice.template.numbering_scheme,
-                        value=invoice.number
-                    ) if invoice else choices.InvoiceNumberingScheme.default_default(),
+                        invoice.number
+                    ) if invoice else provider.invoice_template.numbering_scheme,
```
itsjeyd: @UmanShahzad This will set `number` to the pattern that a provider prefers for invoice numbers (e.g. `"%Y-%m"`), but what we want is an actual number. So I think you'll need to add a `get_default_value` method to `InvoiceNumberingScheme` that works the same way as `increment_value` (calls the appropriate `default_*` method based on the numbering scheme passed in).

accounting/invoice/choices.py
```diff
@@ -36,4 +38,146 @@ class InvoiceTemplate(DjangoChoices):
...
+        Increment the number for the 'opencraft_number' numbering scheme.
+
+        Assumes `value` is in the format of the 'opencraft_number' numbering scheme.
+        """
+        prefix, number = value.split('-', 1)
+        incremented_number = cls.increment_number(number)
+        return '-'.join([prefix, incremented_number])
+
+    @classmethod
+    def increment_value(cls, choice, value):
```
itsjeyd: @UmanShahzad Sorry for not catching this earlier, but I think it would be good to rename the second parameter to `numbering_scheme`; it's a bit more specific than `choice`.

accounting/invoice/utils.py
itsjeyd: @UmanShahzad Ah, good point about bulk payments! Never mind, then. I don't think it would be impossible to get rid of `create_invoice` here and still deal with bulk payments properly. But I see now that this would require some additional conceptual changes to be able to select the right invoice if there's more than one invoice to choose from for a given provider and month. Fine to keep this as is :)

accounting/invoice/choices.py
UmanShahzad: :+1:

accounting/account/models.py
antoviaque: @UmanShahzad Sounds good - that does sound better yep.

accounting/invoice/tests/test_choices.py
itsjeyd: @tomaszgy The question is whether you'd be OK with moving from a numbering scheme that pads numbers with zeroes to a scheme that doesn't? For example, if your last invoice for OpenCraft was `OC-0004`, would you be OK with future invoices being numbered `OC-5`, `OC-6`, etc.?

For context: This PR implements some logic for auto-incrementing invoice numbers according to different schemes. The current version of the code would move you from the scheme involving zeroes to the scheme that doesn't. It would require additional effort to implement an approach that preserves the leading zeroes.

Let me know if that makes sense :)

UmanShahzad: @tomaszgy On your invoices you use OC-0001 as the invoice number. For the next invoice, would you be OK using OC-2, instead of OC-0002? (Trying to use the latter makes things more difficult for the automation code).

UmanShahzad:
@itsjeyd Sorry for the bug, I've fixed that now I believe, and we should be good!

accounting/invoice/views.py
antoviaque: @UmanShahzad I'm not really sure we'll ever have a need for this method? I might be missing it, but I don't see a corresponding user action, since the admin would always be the person marking the invoices as paid?

But that's fine to leave. Just know that on my side, I love to delete code that's not being used :) Even if it is harmless, it's still code we have to maintain - generally, as you have noticed, I would prefer to keep the code simpler and closer to the use case. I've been bitten too many times by code that was implementing features that users didn't actually need, and were sinking time an money that should have gone toward simpler features that users actually really needed - so that's often a large part of my focus in reviews.

UmanShahzad:
@antoviaque Awesome, thanks! I think you've been unintentionally pinging Kshitij here BTW, so don't mind that @xitij2000, unless I'm missing something myself.

accounting/invoice/utils.py
```diff
@@ -21,27 +21,132 @@
...
+            except models.Invoice.DoesNotExist:
+                # This may be the provider's first invoice.
+                invoice = None
+
+            if create_invoice or not invoice:
+                invoice = models.Invoice.objects.create(
+                    number=choices.InvoiceNumberingScheme.increment_value(
+                        invoice.template.numbering_scheme,
+                        invoice.number
+                    ) if invoice else choices.InvoiceNumberingScheme. provider.invoice_template.numbering_scheme,
```
itsjeyd: @UmanShahzad This looks like an "Undo" gone bad? 😄

accounting/invoice/views.py
UmanShahzad: @antoviaque It's a little far-fetched but I can imagine the front-end admin UI having the ability to mark the invoices as paid directly without having to go into Django, and for that we need an endpoint. But I get what you mean -- noted.

accounting/invoice/utils.py
UmanShahzad: :man_facepalming:

antoviaque:
@UmanShahzad 

> I think you've been unintentionally pinging Kshitij here BTW, so don't mind that @xitij2000, unless I'm missing something myself.

Ah, sorry, I meant to ping @itsjeyd , but typing "@ti" puts you first in the list Kshitij. :)

UmanShahzad:
@itsjeyd Thanks a lot for testing this!

So I was a little surprised by your findings, and decided to go through these myself step by step carefully. Here's what I came down to:

1. I did not see the behavior regarding invoices not being sent to the draft folder. They were being sent there, all of them, as expected when calling `send_invoice_approval_request`, and I could not find any invoices in any other folders. When requesting final invoices, they all went to the non-draft folder, as expected. I recommend deleting every single file you can find in the Accounting Stage part of the drive before starting to ensure no confusion there.
1. You expected 2 emails for `send_invoice_prep_notification`, but I saw that both of your emails were of the same domain, i.e. `x@opencraft.com` and `x+a@opencraft.com`. If you change the domain for one of them, you will see that you get emails to both as expected. The reason for not seeing 2 per domain is probably related to using `bcc` vs. `to`. But that doesn't seem to be relevant for our purposes, of course, as long as every unique person is getting the email.
1. The error messages have been around since the beginning IIRC, and I tried looking into them at first but I couldn't find anything of interest and haven't looked deeply again since there were no tracebacks nor misbehavior. My guess (without yet looking) is PyDrive attempts to import some things but falls back repeatedly on older or different versions as necessary. If it was a real issue we'd see a traceback and a crash, but since we haven't, I've ignored it.

itsjeyd:
@UmanShahzad Thanks for the details.

> I did not see the behavior regarding invoices not being sent to the draft folder. They were being sent there, all of them, as expected when calling `send_invoice_approval_request`, and I could not find any invoices in any other folders. When requesting final invoices, they all went to the non-draft folder, as expected. I recommend deleting every single file you can find in the Accounting Stage part of the drive before starting to ensure no confusion there.

To be clear, when testing on stage the behavior was as expected, and I didn't have any problems locating relevant files in the "Accounting Stage" folder on GDrive :) However, my local installation is set up to use the default "Accounting" folder, and when testing there the draft and final invoices ended up in the same subfolder of "Accounting" ("Accounting" > "2018" > "invoices-in" > "01").

I think what might have happened was that the main "Accounting" folder doesn't have a "draft" folder under "2018" > "invoices-in", so the full path for the draft invoice ("Accounting" > "2018" > "invoices-in" > "draft" > "01") doesn't exist. Or is the [code](https://github.com/open-craft/accounting/blob/uman/invoice-approval/accounting/third_party_api/google/mixins.py#L35-L77) for `upload_to_google_drive` supposed to create any missing folders in `target_path`?

Could you check what happens when you connect your local installation to the main "Accouting" folder?

> You expected 2 emails for `send_invoice_prep_notification`, but I saw that both of your emails were of the same domain, i.e. `x@opencraft.com` and `x+a@opencraft.com`. If you change the domain for one of them, you will see that you get emails to both as expected. The reason for not seeing 2 per domain is probably related to using bcc vs. to. But that doesn't seem to be relevant for our purposes, of course, as long as every unique person is getting the email.

Re: both e-mails being of the same domain, isn't that what's going to happen on prod when sending out notifications (assuming that everyone's accounts will be associated with their `@opencraft.com` e-mail address)? Or is "same domain" referring to something else (the `x` portion of the e-mail)?

> The error messages have been around since the beginning IIRC, and I tried looking into them at first but I couldn't find anything of interest and haven't looked deeply again since there were no tracebacks nor misbehavior. My guess (without yet looking) is PyDrive attempts to import some things but falls back repeatedly on older or different versions as necessary. If it was a real issue we'd see a traceback and a crash, but since we haven't, I've ignored it.

Alright, I'm fine with not looking into this further at this stage. But it'd be great to mention it in the documentation (if there's an appropriate place for it), or in an in-line comment, so that others won't need to ask about it if they run into it in the future.

UmanShahzad:
@itsjeyd 

> To be clear, when testing on stage the behavior was as expected, and I didn't have any problems locating relevant files in the "Accounting Stage" folder on GDrive :) However, my local installation is set up to use the default "Accounting" folder, and when testing there the draft and final invoices ended up in the same subfolder of "Accounting" ("Accounting" > "2018" > "invoices-in" > "01").
> I think what might have happened was that the main "Accounting" folder doesn't have a "draft" folder under "2018" > "invoices-in", so the full path for the draft invoice ("Accounting" > "2018" > "invoices-in" > "draft" > "01") doesn't exist. Or is the code for upload_to_google_drive supposed to create any missing folders in target_path?
> Could you check what happens when you connect your local installation to the main "Accouting" folder?

Ah! Ok, yes, this is intentional -- I did not create a `draft` folder in the prod accounting folder, as I was awaiting confirmation of this PR to do that (it's in my notes exactly as such). The code itself is not designed to create new folders if they don't exist, and I think that's okay since our setup is so specific to our own folder structure. I will just create the `draft` folder when this PR merges (will update this PR's 'merge todo').

> Re: both e-mails being of the same domain, isn't that what's going to happen on prod when sending out notifications (assuming that everyone's accounts will be associated with their @opencraft.com e-mail address)? Or is "same domain" referring to something else (the x portion of the e-mail)?

Oh right, I'm using the wrong term: I was basically referring to 2 different email recipients ultimately having their email going to a single email target, i.e. the person given when stripping off any '+a'.

> Alright, I'm fine with not looking into this further at this stage. But it'd be great to mention it in the documentation (if there's an appropriate place for it), or in an in-line comment, so that others won't need to ask about it if they run into it in the future.

Ok, sure, I will mention this in the GDrive client class.

accounting/invoice/tests/test_choices.py
UmanShahzad: @tomaszgy Sorry, I meant yyyy-mm-{number}, not OC-{number} :) 

But yep, we're going to start using it now.

tomaszgy: @UmanShahzad, cool, let'd go ahead with the new schema then :)

PR #11, opened 2018-01-23
Automate TransferWise Bulk Payment CSV Generation

UmanShahzad:
This PR adds an automated job that generates a TransferWise Bulk Payment CSV every month for all unpaid invoices hitherto, and sends that CSV to a pre-set account.

**Testing Instructions**:

1. If necessary, create an account with username `opencraft` (or whatever you set `TRANSFERWISE_BULK_PAYMENT_SENDER` to), with an email you have access to.
1. If necessary, create 1 or 2 provider accounts, and as usual, all of the scaffolding, like hourly rates, invoice templates, and bank accounts.
1. If necessary, run the tasks in `accounting.invoice.tasks`, so that invoices are generated for the providers, and are populated with line items from JIRA.
1. Go ahead and run the task introduced here, in `accounting.transferwise.tasks`.
1. Ensure that the client receives an email with the CSV attached, and that the CSV contains the correct rows and information. Also check that it is uploaded in the appropriate GDrive folder, where the invoices are.


antoviaque:
@UmanShahzad Thank you for this! I appreciate the fact that you did it despite preferring the other approach. I'll let @xitij2000 review, but the code LGTM 👍

UmanShahzad:
@antoviaque Yep, :+1: 

@itsjeyd I have this ready for a code review, and have tested it locally and it matches my expectations. However I didn't get a chance to see the CSV just yet, I'll have to do that on stage once we merge #10. Note that this is based off of #10's PR.

UmanShahzad:
@itsjeyd Update: this is now on stage and I tested it, too. Ready for review :)

accounting/transferwise/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    # Make the new bulk payment for the sender.
+    bulk_payment = models.TransferWiseBulkPayment.objects.create()
+    bulk_payment.create_payments()
+    bulk_payment_csv = bulk_payment.to_bulk_payment_csv(bulk_payment.payments.all())
+
+    # Make the email, attach the CSV and send.
+    email = EmailMessage(
+        subject='TransferWise Bulk Payment CSV',
+        body='TransferWise Bulk Payment CSV is attached.',
```
itsjeyd: @UmanShahzad It would be cool to include the date range that this file covers in the subject and/or body of the e-mail. We can get that info directly from `bulk_payment` so it should be straightforward to include.

accounting/transferwise/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+from django.conf import settings
+from django.core.mail.message import EmailMessage
+from huey import crontab
+from huey.contrib.djhuey import db_periodic_task
+
+from accounting.transferwise import models
+
+
+@db_periodic_task(crontab(day=settings.TRANSFERWISE_BULK_PAYMENT_DAY, hour="0", minute="0"))
+def generate_bulk_payment_csv():
```
itsjeyd: @UmanShahzad It would be nice to continue the naming scheme established in `accounting/invoice/tasks.py`, and call this `send_bulk_payment_csv`.

accounting/transferwise/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Send an email to the designated TransferWise Bulk Payment sender containing the generated bulk payment CSV.
```
itsjeyd: @UmanShahzad Mention that this method also uploads the CSV to GDrive?

```python
    Generate a bulk payment CSV, send it to the designated TransferWise Bulk Payment sender via e-mail,
    and upload it to Google Drive.
```

accounting/transferwise/tests/test_tasks.py
```diff
@@ -0,0 +1,92 @@
...
+@ddt.ddt
+class GenerateBulkPaymentCsvTestCase(TestCase):
+    """ Test cases for `tasks.generate_bulk_payment_csv`. """
+
+    @freezegun.freeze_time(NOW)
+    def setUp(self):
+        """ Set up test objects. """
+        self.create_client_and_provider_links()
+
+        # The providers need valid bank accounts  to get paid.
```
itsjeyd: @UmanShahzad Typo: Two spaces between "accounts" and "to".

accounting/transferwise/tests/test_tasks.py
```diff
@@ -0,0 +1,92 @@
...
+        # Create some unpaid invoices that need to be paid in bulk.
+        self.invoice1 = factories.InvoiceFactory(provider=self.provider1, client=self.client, paid=False)
+        self.invoice2 = factories.InvoiceFactory(provider=self.provider2, client=self.client, paid=False)
+        super().setUp()
+
+    @mock.patch.object(models.TransferWiseBulkPayment, 'upload_to_google_drive')
+    @mock.patch.object(EmailMessage, 'attach_file')
+    @mock.patch.object(models.TransferWiseBulkPayment, 'to_bulk_payment_csv')
+    def test_generate_bulk_payment_csv(self, mock_to_bulk_payment_csv, mock_attach_file, mock_upload_to_google_drive):
+        """ A bulk payment CSV is generated for all unpaid invoices and sent off to the bulk payment sender. """
```
itsjeyd: @UmanShahzad Same note here about mentioning the upload to GDrive.

```python
        """ A bulk payment CSV is generated for all upaid invoices, sent of to the bulk payment sender, and uploaded to Google Drive. """
```

accounting/transferwise/tests/test_tasks.py
```diff
@@ -0,0 +1,92 @@
...
+        # 1 sender, so 1 call to everything.
+        mock_to_bulk_payment_csv.assert_called_once()
+        mock_attach_file.assert_called_once()
+        mock_upload_to_google_drive.assert_called_once()
+        self.assertEqual(len(mail.outbox), 1)
+
+        # 1 bulk payment that contains 2 payments, 1 for each provider.
+        self.assertEqual(models.TransferWiseBulkPayment.objects.all().count(), 1)
+        self.assertEqual(models.TransferWisePayment.objects.all().count(), 2)
+        self.assertEqual(models.TransferWiseBulkPayment.objects.first().payments.all().count(), 2)
```
itsjeyd: @UmanShahzad I think you can simplify the code above by calling `count()` on the manager (instead of calling `all()` first). Also, this test expects there to be exactly one `TransferWiseBulkPayment`, so we could use `get` instead of `first`:

```python
        self.assertEqual(models.TransferWiseBulkPayment.objects.count(), 1)
        self.assertEqual(models.TransferWisePayment.objects.count(), 2)
        self.assertEqual(models.TransferWiseBulkPayment.objects.get().payments.count(), 2)
```

Makefile
```diff
@@ -21,7 +21,7 @@
 
 .DEFAULT_GOAL := help
 HELP_SPACING ?= 30
-COVERAGE_THRESHOLD ?= 80
+COVERAGE_THRESHOLD ?= 81
```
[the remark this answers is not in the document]
UmanShahzad: :+1: Yea!

accounting/transferwise/migrations/0003_farewell_sender.py
```diff
@@ -0,0 +1,31 @@
...
+        ),
+        migrations.RemoveField(
+            model_name='transferwisebulkpayment',
+            name='auto_upload_google_drive_on_save',
+        ),
+        migrations.RemoveField(
+            model_name='transferwisebulkpayment',
+            name='sender',
+        ),
+    ]
```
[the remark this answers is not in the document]
UmanShahzad: Lul :)

accounting/transferwise/models.py
```diff
@@ -114,31 +104,16 @@ def create_payments(self):
         TODO: we could then skip generating CSVs and create all the transfers in one go, and call them a 'bulk payment',
         TODO: even if that terminology is only to represent how we're doing it internally.
         """
-        unpaid_invoices = self.sender.client_invoices.filter(paid=False, date__range=(self.start_date, self.end_date))
+        unpaid_invoices = self.sender.client_invoices.filter(
+            paid=False,
+            date__range=(self.start_date, self.end_date),
+            provider__bank_accounts__transferwise_recipient_id__isnull=False,
```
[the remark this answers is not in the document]
UmanShahzad: I could be cleaner and make a custom query manager, and that might actually help in other places in the codebase actually, but I've just chosen not to at the moment. Might make a small ticket to update that across the codebase if we have time for it. (Someone else could do it, like how Daniel's doing the GDrive link ticket atm).

accounting/transferwise/tasks.py
UmanShahzad: That's probably a good idea, yeah.

accounting/transferwise/tasks.py
UmanShahzad: :+1:

accounting/transferwise/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    # Make the new bulk payment for the sender.
+    bulk_payment = models.TransferWiseBulkPayment.objects.create()
+    bulk_payment.create_payments()
+    bulk_payment_csv = bulk_payment.to_bulk_payment_csv(bulk_payment.payments.all())
```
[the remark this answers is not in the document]
UmanShahzad: PyCharm complains and says 'payments' isn't part of 'self' or something in the mixin class, which is annoying to see, but also technically true. But yeah, since we're still using the `TransferWisePayment` class' API in the mixin (by saying `payment.invoice`), it probably just makes sense to assume connection to something that has `payments` as a variable. So let's change it.

accounting/transferwise/tasks.py
UmanShahzad: We actually plan on making the bulk payment's date range be from ancient history to now! That means we'd always just get "super old date to now" as the date range. The only way to tell appropriately, is to check the bulk payment's individual payments, and get the min and max of date ranges for all the invoices in them. :p

accounting/transferwise/tests/test_tasks.py
UmanShahzad: :+1:

accounting/transferwise/tests/test_tasks.py
UmanShahzad: :+1:

accounting/transferwise/tests/test_tasks.py
UmanShahzad: Oh yes, good catches :+1:

accounting/transferwise/models.py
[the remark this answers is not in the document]
itsjeyd: @UmanShahzad Sounds good, thanks for the details.

accounting/transferwise/tasks.py
itsjeyd: @UmanShahzad Ah, OK :) Maybe the date and time at which the CSV was generated would still be interesting? Feel free to consider this nit-level/polish, though -- no need to keep iterating here to address this.

UmanShahzad: @itsjeyd That would be an option, yeah, but the date the CSV is generated is the same as when the email was sent, since it all happens in one go, so the timestamp of the email would give the same information.

UmanShahzad:
@itsjeyd Thanks!!

UmanShahzad:
Rebased on `master` from 41c7475b131fdc142ea119627b61dc5bc3b23dbb

UmanShahzad:
@antoviaque Thanks!

PR #12, opened 2018-01-26
Use alternateLink for the PDF to avoid 400 errors

clemente:
When creating a PDF invoice, `webContentLink` provides a link that will show a 403 error when the user is signed in from a different account. The new link, `alternateLink`, can show an account chooser to switch to another account.


UmanShahzad:
@clemente Thanks! Don't mind the failing tests, BTW, there's some really weird flaky thing going on that I've been having to deal with for a while now. #11 will solve this. For now the only way to fix it is by restarting tests without cache until it eventually works. Don't bother -- I'll do it at the end before merging.

Will test this soon.

PR #15, opened 2018-01-28
Use Pipfile and pipenv.

UmanShahzad:
This PR uses the new pipfile/pipenv scheme of dependency and virtual environment management.

**Testing Instructions**:

* Ensure this works: `make dev.down` -> `make dev.build` -> `make dev.up` -> `make dev.web.logs` & `make shell` & `curl http://localhost:1786/admin`.
* Ensure this works: `make dev.down` -> `vagrant up --provision` -> `vagrant ssh` -> `make run.dev` & `make shell` & `curl http://localhost:1786/admin`.

The complementary ops PR for this is at https://github.com/open-craft/ansible-accounting/pull/2.


Pipfile
```diff
@@ -0,0 +1,142 @@
...
+
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+alabaster = "==0.7.10"
+"asn1crypto" = "==0.24.0"
```
itsjeyd: @UmanShahzad This is the first time I'm learning about `pipfile`. Why wrap some package names with quotes but not all of them?

bin/setup
```diff
@@ -1,39 +1,31 @@
...
 
-# Need python-3.6 repository
+# Need python-3.6 repository.
 echo "Setting up python3.6 repository"
 sudo add-apt-repository 'ppa:deadsnakes/ppa'
-sudo apt-get -y update
+sudo apt-get update
 
-# Install all dependencies.
+# Install all system and python dependencies.
```
itsjeyd: @UmanShahzad "... and activate this project's virtualenv."? (via `pipenv shell` below)

bin/setup
```diff
@@ -47,14 +39,7 @@ host    all             all             127.0.0.1/32            trust
...
-
-# Need to upgrade pip 
-pip install --upgrade setuptools
 
-# Setup Python Requirements
-cd ~/accounting
-pip install -r requirements.txt
-
-# Initial Migration
+# Begin DB services with migrations finished.
```
itsjeyd: @UmanShahzad This is a little confusing: The code below restarts `postgresql` before running migrations?

Pipfile
```diff
@@ -0,0 +1,142 @@
...
+"urllib3" = "==1.22"
+virtualenv-clone = "==0.2.6"
+virtualenv = "==15.1.0"
+wcwidth = "==0.1.7"
+websocket-client = "==0.45.0"
+Werkzeug = "==0.13"
+wrapt = "==1.10.11"
+
+
+[dev-packages]
```
itsjeyd: @UmanShahzad Do we need to add this now if we're not specifying any packages below?

UmanShahzad:
Rebased on `uman/improved-admin` from 8d46cb31441c85462e8514535b984e7c3e74802a.

Pipfile
UmanShahzad: @itsjeyd It turns out `pipenv install` generates a `Pipfile` for you if it doesn't exist but a `requirements.txt` does (which is probably the main way they'll get masses to switch to it eventually), so this was generated by it. The pattern shows that quotes are around any name with a number in it, like asn**1**crypto, etc..

Pipfile
UmanShahzad: @itsjeyd Since it was generated, and I didn't even know it existed, it might be good to leave around in case we decide to add stuff there (I can see that happening even now, but as a first time with Pipfile and considering things could change with their implementation, I'm not messing further with it except what was generated for me).

bin/setup
UmanShahzad: @itsjeyd :+1:

bin/setup
UmanShahzad: @itsjeyd Yeah that comment is pretty bad. I think I should just delete the comment, the commands are very self-descriptive.

itsjeyd: @UmanShahzad Nice :)

UmanShahzad:
@itsjeyd For docker, there did seem to be some issues, yep. I've updated the Dockerfile to reflect the proper way of handling it. Just do `docker rmi python:3.6` (if that's the only python3.6 image you have), and `make dev.up` will re-download and provision, and it should work this time.

For vagrant, I'm kinda stuck because I can't test it on my local machine: ever since I applied the security patches my vagrant has been failing to work, in the beginning it was hard-crashing my computer. I'll see what I can do.

And yep, I think I need to update the README, thanks for mentioning that!

UmanShahzad:
@itsjeyd Hmm, that's very weird. I think we may have different versions of vagrant and/or the box that's causing this. Before I updated my vagrant/virtualbox versions to fix my problems with vagrant, I also was getting logged in as `ubuntu`, but later was going in as `vagrant`. It turns out the same happens on the Ocim box (you get in as `vagrant`).

Not entirely sure what's the way forward. Could you try doing this with a completely fresh box? I.e. removing the current one and then pulling and provisioning anew? Either that would fix it (I hope?), or we'll have to change `bin/setup` to force-use the `vagrant` user or something.

PR #16, opened 2018-02-01
Allow only one bank account per user account

clemente:
This changes BankAccount from ForeignKey to OneToOneField so that each user account can have only one bank account, not many. We still don't have use cases to allow multiple bank accounts per user.

To review this
- run tests (though right now I get 8 failures for `January`/`December`, and 1 for a missing `InvoiceTemplate` object, both unrelated)
- check the code
- test in admin: add, delete
- run `generate_account_data`
- maybe check API (though now it's failing with `'InvoiceViewSet' should either include a `serializer_class` attribute`)


UmanShahzad:
@clemente Can you rebase on master now? The tests are fixed :)

accounting/account/serializers.py
```diff
@@ -44,7 +44,7 @@ class AccountSerializer(UuidModelSerializer):
     class Meta(UuidModelSerializer.Meta):
         model = models.Account
         fields = (UuidModelSerializer.Meta.fields +
-                  ('user', 'address', 'business_name', 'occupation', 'bank_accounts', 'vat',))
+                  ('user', 'address', 'business_name', 'occupation', 'bank_account', 'vat',))
```
UmanShahzad: You have to update the `bank_accounts` above to be `bank_account` too.

UmanShahzad: @clemente Shoot, one more change I had to do locally: the serializer cannot use `many=True` since it's just 1-to-1 now.

clemente: Both fixed, thanks.

clemente:
Now the tests pass in local!

PR #17, opened 2018-03-08
Fix up some links that were still inaccessible.

UmanShahzad:
@antoviaque I checked and noticed TransferWise bulk payment CSV links in the billing admin were inaccessible. Indeed, it looks like we may have either missed or accidentally added back the wrong key for getting the link.

See https://github.com/open-craft/accounting/pull/12 which had originally fixed this.

*Testing instructions*:

1. Go to production to see the existing TransferWise bulk payments and notice you cannot view the CSV files by clicking on the link in the admin view. Google just gives you a 4xx page.
1. On stage, which has this deployed, follow (roughly) the logic in https://github.com/open-craft/accounting/blob/master/accounting/transferwise/tasks.py#L33 to create the bulk payment, create the CSV version, and upload to Google Drive (you can skip the email bit of course).
1. Go into the billing console and see the CSV link for the new bulk payment you created. Click it and notice that it works without 4xx errors.

### DIFF
--- a/accounting/transferwise/tasks.py
+++ b/accounting/transferwise/tasks.py
@@ -53,5 +53,5 @@ def send_bulk_payment_csv():
         target_path=[bulk_payment.date.strftime('%Y'), 'invoices-in', bulk_payment.date.strftime('%m')],
         title=bulk_payment.csv_filename,
     )
-    bulk_payment.csv_path = file['webContentLink']
+    bulk_payment.csv_path = file['alternateLink']
     bulk_payment.save()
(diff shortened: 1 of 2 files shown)


UmanShahzad:
@antoviaque Sure, we can have it tested first.

@itsjeyd Do you think you'd have time to check this out? It's a pretty straightforward change as you can tell. Let me know and I'll write testing instructions.

itsjeyd:
@UmanShahzad Yup, I can review/test this :)

UmanShahzad:
@itsjeyd Alrighty, thanks! Testing instructions are updated in the main post, and this is on stage.

UmanShahzad:
@itsjeyd Thanks for reviewing so quickly!

